Issue #16, opened 2020-02-28
Failure when writing field with converted datatype


davidhassell:
When converting the data type of an output array with the `datatype` keyoword of `cfdm.write`, the `_FillValue` is not converted, leading to a netCDF error:
```
>>> import cfdm
>>> f = cfdm.example_field(1)
>>> f.set_property('_FillValue', 45.)
>>> cfdm.write(f, 'delme.nc', datatype={numpy.dtype('float64'): numpy.dtype('float32')})
<snip>
AttributeError: NetCDF: Not a valid data type or _FillValue type mismatch
```

The solution is simply to make sure that `_FillValue` and `missing_value` attributes are converted if required.

davidhassell:
This will be fixed in version 1.8.0.

PR #17, opened 2020-02-28
FillValue dtype match

davidhassell:
issue #16


davidhassell:
This fix will go into v1.8.0 instead

PR #29, opened 2020-04-27
Warnings when writing valid_[min|max|range] properties. Fixes #30

### LINKED ISSUE #30: Issue a warning if reading and writing datasets that have valid_[min|max|range] attributes
The presence of the the `valid_min`, `valid_max` or `valid_range` attributes causes data to be masked for its "out-of-range" values. This is sometime surprising, particularly if the data has been modified between the read and write operations.

**Proposal:**

Add a `warn_valid` keyword parameter, default `True` to `cfdm.read` and `cfdm.write` that warns when such properties are present. The `cfdm.write` case will only warn if the data contains out-of-range values. Since data is not actually read by `cfdm.read` (lazy loading), it is not possible to check for out-of-range data during the read, so merely the presence of any of these properties will trigger the warning.

If `warn_valid=False` then warnings are suppressed.

davidhassell:


davidhassell:
Hi @sadielbartholomew It'd be great if you could review this PR. We discussed have `cfdm.write` raise an exception, rather than printing a warning, but I've had second thoughts.  I think that it goes against the grain to fail on something that is, after all, CF-compliant.

davidhassell:
Fixes #30

sadielbartholomew:
> It'd be great if you could review this PR.

Sure. I've assigned myself so I can get notified automatically when it is ready for review.

> We discussed have cfdm.write raise an exception, rather than printing a warning, but I've had second thoughts. I think that it goes against the grain to fail on something that is, after all, CF-compliant.

Fair enough, that does sound sensible. A warning is indeed sufficient & if users don't read it & heed it then that is their problem!

sadielbartholomew:
(Holding on reviewing for the moment as it seems from external discussion we might want to adjust the approach.)

davidhassell:
Further to offline discussions:

1\) I think is OK, as the data being written is within the valid range. Assigning a large (+ve or -ve) number to the data array prior to writing triggers the warning for me.

2\) has been hopefully been fixed by c112ed2 and 7643d4f

@sadielbartholomew Can you confirm? Thanks

sadielbartholomew:
Actually since you may be working on other things, I'll go in & see if I can work out what needs to be tweaked... I will share a ``diff`` if so.

davidhassell:
Hmm. Works for me. Perhaps I hadn't pushed up the relevant changes (sorry)? Try 6f229e5 ...

sadielbartholomew:
> Hmm. Works for me. Perhaps I hadn't pushed up the relevant changes (sorry)? Try 6f229e5 ...

Strange, I have checked out your updated branch & it the written file has still gone AWOL. If I run it with ``verbose=True`` it shows it is being written fine. Maybe I'm doing something silly, let me take a closer look at the lines I am using...

sadielbartholomew:
Aha! It is writing them, but to the current directory (where I am running the iPython shell), rather than into ``~/cf-test-datasets/``, which is where it was writing them before. Actually, the current write location seems the correct one. I wonder why it was writing to the latter previously in this PR?

sadielbartholomew:
Darn it, there are some conflicts to resolve. I shall try to resolve them now...

sadielbartholomew:
Conflicts resolved. All resolutions were obvious & most conflicts were caused by my relatively recent addition of the in-place operations decorators.

PR #49, opened 2020-06-18
Address all user-facing TODOs & cut duplicate class

sadielbartholomew:
Fill in all TODOs that [are currently rendered in the documentation](https://ncas-cms.github.io/cfdm/search.html?q=TODO).

Additionally:

* remove the class ``NetCDFHDF5_exp`` as it is not documented or used, & the methods in there are almost duplicates (except with names ending ``_chunksize`` instead of the plural) in purpose to those in documented class above, ``NetCDFHDF5``, from which I have addressed many TODOs here (@davidhassell is that okay? I suspect that it was an earlier version of the near-equivalent documented class?)
* change a small(er) number of uses of "chunk size" to "chunksize" in text content to be consistent (I have seen it written both ways so either is fine but the latter was more common in docstrings in the codebase).


sadielbartholomew:
I have rebuilt the latest documentation to pick up on the changes & demonstrate there are no TODOs left within, but can drop it & rebuild if any changes are suggested. Check the first commit changes for all the addressing of the TODOs without all the docs build files cluttering it up.

sadielbartholomew:
I have just removed the final docs build commit as we agreed before I went on leave it is best to only re-build at release-time.

sadielbartholomew:
I've just resolved the conflicts that had emerged, none of them being controversial.

There's no point checking he CI tests as at this stage they will fail by default because they now require the ``netcdf_flattener`` package & it is not available on pip or conda (quite yet, as I understand it) so would need some manual steps added to the workflow to install it. However I've installed the netCDF Flattener locally & all tests for this revised branch pass. Therefore I am going to merge.

PR #69, opened 2020-08-13
New write mode to append to existing netCDF file

### LINKED ISSUE #143: Appending data to an existing NetCDF file
See NCAS-CMS/cf-python#30 for details (opening this Issue mainly as a way to elegantly reference the issue at hand in the cfdm Changelog).

sadielbartholomew:
Closes #143.

WIP for now. Details to follow once the PR is ready-for-review.


sadielbartholomew:
Hi @davidhassell, as promised in order to check I am tying up the remaining loose ends correctly, here is my interpretation of the complications we perceived and corresponding decisions we converged upon after the discussions yesterday.

I am in the process of finalising implementation of the following decisions and also of other concluding tasks such as updating the documentation in the way we discussed and putting in the PR to provide means to use this feature in cf-python. Therefore please let me know ASAP if there is anything that you think isn't right, or has been missed, or indeed if you have had other thoughts since!

******************

### Problem space

#### Complications to consider summarised:
1. Domain ancillaries: formula terms for these are not (either always or ever, not sure with only one tested thus far) being interpreted properly, leading to an extra field being created for every field having any domain ancils (e.g. as demonstrated by appending the `n=1` `example_field` in the current unit test).
2. Global attributes (generally): apply to everything in the file, so what rule should govern the global attributes that will exist on the final, post-append, file given the set of global attributes on the original field and on any appended fields?
3. `featureType` as a special global attribute case: it has structural meaning and only one per file makes sense, but the user could try to append fields with different `featureType`s either to themselves or to the fields in the original file. 
4. Group attributes: groups have to work for whole file, so it is a similar case to global attributes, but unlike those group ones can be nested so are potentially more complicated to handle.

#### General philosophy to guide decisions

* To "do no harm".
* To not edit fields that exist in the original (pre-append) file at all: fields already encoded should be preserved.

### Decisions made

With respect to each item in the 'Complications' list we agreed to the following as I understood.

1. Set and document as a known limitation for the initial append mode (i.e. the one of this PR) that will go into the imminent release and so to open an Issue on this with the aim to investigate the true cause and resolve it for the next release, probably 1.8.9.1. Still allow fields with domain ancils to be appended, but warn that they could be set on
    * as for the documenting, I am thinking to put a fairly generic sentence into the mode parameter part of the `write` docstring as well as adding a log message at warning level when an append-mode write is done where any appended field is detected to have domain ancils.

2. As a strict rule we will not allow the global attributes to change on the original file i.e. the output file global attributes are the input global file attributes, regardless of whatever they are for the appended fields.
    * we realised this could probably be implemented simply, by protecting the `write_global_attributes' calls with an `if` statement testing that it the logic is not running in append mode.

3. As a case-by-case rule: if there is a `featureType` set on the original file, all appended fields must have that same `featureType`, otherwise (without one set on the original file), all appended fields must also not have a `featureType`.
    * the unit test will therefore need some special checks for `featureType`, to be added shortly;

4. Set and document as a known limitation (as with (1)) whereby we forbid (via `ValueError`), for this `v1.8.9.0` append mode, any fields with groups (other than the "root" group) from being appended, regardless of whether there are groups within the original file. Like for (1) we are aiming to agree a proper approach (one that sounded sensible was outlined) and remove this limitation for the next release.
    *  it would be useful towards this to create a new utility in `cfdmimplementation` that gets all of the groups associated with a field (I will double-check there isn't one already we weren't aware of).


************

davidhassell:
Hi @sadielbartholomew - thanks - I think that all these points are good.

sadielbartholomew:
Just adding some issues to cover the agreed known limitations in advance so I can link to them in the relevant commits...

sadielbartholomew:
Hi @davidhassell, I am afraid I still have the global attributes, `featureType`, docs and Changelog aspects (only) left to complete, largely because I hit a bit of difficulty with keeping the global attributes the same, in fact I now think we have to do a little more than just protect `_write_global_attributes` calls to get what we want, e.g. based on the docs:

> If any field construct’s property has a different value then the property will not be written as a netCDF global attribute, even if it has been selected as such, but will appear instead as attributes on the netCDF data variables corresponding to each field construct.

but in the interests of time to get a release out ready for Monday, and given that those omissions are quite self-contained, I am opening this for review now. I will finish those outstanding aspects tomorrow as soon as I "clock in" so please review what is here. In particular, it would be useful immediately to hear if you are seeing any seg faults towards the end of the final netCDF4 iteration, since I am intermittently seeing those but only ever in that context now. Thanks.

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -4723,23 +4901,18 @@ def write(
         # ------------------------------------------------------------
         # Open the output netCDF file
         # ------------------------------------------------------------
-        filename = os.path.expanduser(os.path.expandvars(filename))
-        if os.path.isfile(filename):
+        if os.path.isfile(filename) and mode == "w":
```
davidhassell, line 4904: I think the logic here is a bit awry, as _not_ setting overwrite=False when the file does not exist causes problems in `def file_open`

Replacing this line, and the following two if blocks, with
```python
        if os.path.isfile(filename):
            if mode == "w" and not overwrite:
                raise IOError(
                    "Can't write with mode {mode!r} to existing file "
                    f"{os.path.abspath(filename)} unless overwrite=True"
                )

            if not os.access(filename, os.W_OK):
                raise IOError(
                    "Can't write to existing file "
                    f"{os.path.abspath(filename)} without permission"
                )
        else:
            g["overwrite"] = False
```
should do it ...

cfdm/read_write/write.py
```diff
@@ -165,6 +166,22 @@ def write(
             ``'NETCDF4'`` files use the version 4 disk format (HDF5)
             and use the new features of the version 4 API.
 
+        mode: `str`, optional
```
davidhassell, line 169: I think we need a bit of white space to make the table render correctly:

```python
        mode: `str`, optional
            Specify the mode of write access for the output file. One
            of:

            =======  =================================================
            *mode*   Description
            =======  =================================================
            ``'w'``  Open a new file for writing to. If it exists and
                     *overwrite* is True then the file is deleted
                     prior to being recreated.

            ``'a'``  Open an existing file for appending new
                     information to. The new information will be
                     incorporated whilst the original contents of the
                     file will be preserved.
            =======  =================================================

            By default the file is opened with write access mode
            ``'w'``.

```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -4595,7 +4657,131 @@ def write(
...
+            # and 'a' for dry_run and post_dry_run respectively) so that the
+            # mode does not need to be passed to various methods, where a
+            # pair of such keys seem clearer than one "effective mode" key.
         }
+
+        effective_mode = mode  # actual mode to use for the first IO iteration
+        effective_fields = fields
+        if mode == "a":
+            # First read in the fields from the existing file:
+            effective_fields = NetCDFRead(_implementation).read(filename)
```
davidhassell, line 4679: Better to use self.implementation, i.e. effective_fields = NetCDFRead(self.implementation).read(filename), as this will play nicely with downstream applications (i.e. cf.python)
```suggestion
            effective_fields = NetCDFRead(self.implementation).read(filename)
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -10,11 +10,16 @@
 
 from .. import IOWrite
 
+from .netcdfread import NetCDFRead
+
+from ...cfdmimplementation import implementation
```
davidhassell, line 15: Better to use `self.implementation`, i.e.  `effective_fields = NetCDFRead(self.implementation).read(filename)`, as this will play nicely with downstream applications (i.e. cf.python) 
```suggestion
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -10,11 +10,16 @@
...
 from ...decorators import _manage_log_level_via_verbosity
 
 
 logger = logging.getLogger(__name__)
 
+_implementation = implementation()
```
davidhassell, line 21: Better to use self.implementation, i.e. effective_fields = NetCDFRead(self.implementation).read(filename), as this will play nicely with downstream applications (i.e. cf.python)
```suggestion
```

cfdm/test/test_read_write.py
```diff
@@ -172,10 +177,174 @@ def test_read_write_format(self):
...
+                # Note: after Issue #141, this skip can be removed.
+                if ex_field_n == 1:
+                    continue
+                # Skip since "Can't write int64 data from <Count: (2) > to a
+                # NETCDF3_CLASSIC file" causes a ValueError i.e. not possible.
+                # Note: can remove this when Issue #140 is closed.
+                if fmt in self.netcdf3_fmts and ex_field_n == 6:
+                    continue
+
+                cfdm.write(ex_field, tmpfile, fmt=fmt, mode="a")
```
davidhassell, line 207: I'm getting a failure here for `fmt=NETCDF4_CLASSIC` and `n=3`:

```
  File "/home/david/cfdm/cfdm/read_write/netcdf/netcdfwrite.py", line 270, in _write_attributes
    g["nc"][ncvar].setncatts(netcdf_attrs)

AttributeError: NetCDF: Attempt to define fill value when data already exists.
```

This looks to me like an unrelated problem getting trapped at this low level trap. Do you see this?

The field is a DSG - could that be part of it? although I can't see how that would affect things in this way ....

cfdm/read_write/netcdf/netcdfwrite.py
sadielbartholomew: Ah, good spot and thanks for the suggested fix, I'll give it a try in the next commit and report back.

cfdm/read_write/write.py
sadielbartholomew: Another good spot, thanks. I'll add that in as part of the documentation commit. Generally once I've made and am happy with the docs changes I planned to go in and make sure they rendered correctly but it's good to know this in advance.

cfdm/read_write/netcdf/netcdfwrite.py
sadielbartholomew: Excellent suggestion again, thanks. Not sure how I missed that since it is already used throughout the class!

I'll commit your suggestion here but also remove the next commit the explicit `implementation` import and variable definition I added previously related to this and they will no longer be required (as a bonus).

cfdm/read_write/netcdf/netcdfwrite.py
sadielbartholomew: Oops, ignore my comment on the previous item saying I'll remove this in the next commit, you've covered it. Thanks.

cfdm/test/test_read_write.py
sadielbartholomew: Ah, important find there, good job; I have seen that one before, but thought recent changes had fixed it as I hadn't seen it for a while.

The error message would indeed suggest an unrelated problem, though regardless of whether or not it is related to this new code we should get it fixed for the imminent release, I think. I'm looking into it...

cfdm/read_write/netcdf/netcdfwrite.py
sadielbartholomew: The suggestion works perfectly, thanks, and has been added in one of the new commits I will push shortly.

sadielbartholomew:
The first set of review feedback should all be addressed now, except the final comment left unresolved above. I've been working in parallel on addressing the outstanding tasks.

sadielbartholomew:
By enabling this feature upstream in cf-python (PR to follow) I have noticed a further potential issue left to resolve, namely that fields with a similar structure are not being appended, e.g. (in cf-python) the latter field is not being appended where the former is present:

```pycon
>>> import cf
>>> print(cf.example_field(2))
Field: air_potential_temperature (ncvar%air_potential_temperature)
------------------------------------------------------------------
Data            : air_potential_temperature(time(36), latitude(5), longitude(8)) K
Cell methods    : area: mean
Dimension coords: time(36) = [1959-12-16 12:00:00, ..., 1962-11-16 00:00:00]
                : latitude(5) = [-75.0, ..., 75.0] degrees_north
                : longitude(8) = [22.5, ..., 337.5] degrees_east
                : air_pressure(1) = [850.0] hPa
>>> print(cf.example_field(5))
Field: air_potential_temperature (ncvar%air_potential_temperature)
------------------------------------------------------------------
Data            : air_potential_temperature(time(118), latitude(5), longitude(8)) K
Cell methods    : area: mean
Dimension coords: time(118) = [1959-01-01 06:00:00, ..., 1959-02-28 18:00:00]
                : latitude(5) = [-75.0, ..., 75.0] degrees_north
                : longitude(8) = [22.5, ..., 337.5] degrees_east
                : air_pressure(1) = [850.0] hPa
```

sadielbartholomew:
Hi @davidhassell, all pushed except the ``featureType`` approach implementation which led to some hanging for some reason so I am revising. Won't be long and I will commit and push that.

Please can you test this on your env and let me know if you are still seeing https://github.com/NCAS-CMS/cfdm/pull/69#discussion_r632407424? And if you like you could review the commits made since your initial review, etc. Thanks.

sadielbartholomew:
Opening and closing to trigger the Actions jobs now...

sadielbartholomew:
(Previous feedback discussed and further feedback raised externally to GitHub, on Slack.)

The CI jobs and Codacy are happy! Saying that, they didn't pick up on one issue I noticed as soon as I started those jobs (naturally), where a flag variable was defined in slightly the wrong place. The following commit fixes that.

All feedback should now be addressed, what do you think @davidhassell ? 

I won't merge this ye, though, since I have a few minor issues to resolve that only show up in the equivalent cf-python test.

cfdm/test/test_read_write.py
sadielbartholomew: Should now be fixed...

sadielbartholomew:
https://github.com/NCAS-CMS/cfdm/pull/69#issuecomment-842477633 is not an issue I have realised as for the `cf` example fields, `n=5` and `n=2` are aggregatable and indeed aggregate down to one field:

```pycon
>>> import cf
>>> b = cf.example_field(2)
>>> print(b)
Field: air_potential_temperature (ncvar%air_potential_temperature)
------------------------------------------------------------------
Data            : air_potential_temperature(time(36), latitude(5), longitude(8)) K
Cell methods    : area: mean
Dimension coords: time(36) = [1959-12-16 12:00:00, ..., 1962-11-16 00:00:00]
                : latitude(5) = [-75.0, ..., 75.0] degrees_north
                : longitude(8) = [22.5, ..., 337.5] degrees_east
                : air_pressure(1) = [850.0] hPa
>>> c = cf.example_field(5)
>>> print(c)
Field: air_potential_temperature (ncvar%air_potential_temperature)
------------------------------------------------------------------
Data            : air_potential_temperature(time(118), latitude(5), longitude(8)) K
Cell methods    : area: mean
Dimension coords: time(118) = [1959-01-01 06:00:00, ..., 1959-02-28 18:00:00]
                : latitude(5) = [-75.0, ..., 75.0] degrees_north
                : longitude(8) = [22.5, ..., 337.5] degrees_east
                : air_pressure(1) = [850.0] hPa
>>> c.equals(b, verbose=-1)
Data: Different shapes: (118, 5, 8) != (36, 5, 8)
Field: Different data
False
>>> a = cf.aggregate([b, c])
>>> a
[<CF Field: air_potential_temperature(time(154), latitude(5), longitude(8)) K>]
```

which (noting the `154` rather than `118` time points) is the field indeed set in the file post-append, so the test just needs adapting to test that `n=5` doesn't change the length of the overall field list. I am about to push a commit that does that over on NCAS-CMS/cf-python#213.

sadielbartholomew:
(Discussing via Slack and investigating some slowness emerging in another `test_read_write` test before we are confident to that this is good to merge.)

sadielbartholomew:
https://github.com/NCAS-CMS/cfdm/pull/69#issuecomment-846178520 turned out to be unrelated to the PR code changeset and has been addressed in 
87c8c5a, with all credit to @davidhassell :smiley: Now we should be ready to go...

davidhassell:
All tests pass for me, and in the expected amount of time (apart from a pep8 white space error in `test_read_write.py` - but that's probably just me, as I had to fix a merge conflict in there).

Good work!

davidhassell (approved):

sadielbartholomew:
Opening-closing to check the Actions jobs still pass...

sadielbartholomew:
:tada: :balloon:  Good to merge on all counts! Thanks for all of your help with this @davidhassell, we got there in the end... Now to the cf-python equivalent; review to be opened on #NCAS-CMS/cf-python#213 very soon.

PR #70, opened 2020-08-20
Docstring rewriter

davidhassell:
Allow reuse of docstrings in different classes, and of docstring snippets.


sadielbartholomew (commented):
I'm mid-review, and generally this is looking fantastic and working a treat as far as I have observed, but I have a few comments that I will register up front:

1. When calling the help function on classes, some of the new dunderscore docstring methods are showing up. I don't think we should expose the user to those (nobody except us poor developers!) unless you were aware they were appearing and thought otherwise?

    Example:

    ```console
    $ help(cfdm.AuxiliaryCoordinate)
    Help on class AuxiliaryCoordinate in module cfdm.auxiliarycoordinate:

    class AuxiliaryCoordinate(cfdm.mixin.netcdf.NetCDFVariable, cfdm.mixin.coordinat
    e.Coordinate, cfdm.core.auxiliarycoordinate.AuxiliaryCoordinate)
     |  AuxiliaryCoordinate(properties=None, data=None, bounds=None, geometry=None, 
    interior_ring=None, source=None, copy=True, _use_data=True)
     |  
     |  An auxiliary coordinate construct of the CF data model.
    ...
    ...
    ...
     |  Methods defined here:
     |  
     |  __deepcopy__(self, memo)
     |      Called by the `copy.deepcopy` function.
     |      
     |      x.__deepcopy__() <==> copy.deepcopy(x)
     |      
     |      .. versionadded:: (cfdm) 1.7.0
     |      
     |      **Examples:**
     |      
     |      >>> import copy
     |      >>> y = copy.deepcopy(x)
     |  
     |  __docstring_package_depth__(self)
     |      Return the package depth for cfdm docstring substitutions.
     |      
     |      See `_docstring_package_depth` for details.
     |  
     |  __docstring_substitutions__(self)
     |      Define docstring substitutions that apply to this class and all of
     |      its subclasses.
    ...
    ...
    ...
    ```

2. In the new `rewritedocstring` module there seems to be a lot of use of `for` loops equivalent logically (using `a`, `b` etc for dummy objects) to:

    ```python
    for a in b[::-1]:
        c = getattr(a, _<some docstring method basename>, None)
        if c is not None:
            d.update(c(a))
        else:
            c = getattr(a, __<some docstring method basename>__, None)
            if c is not None:
                d.update(c(None))
    ```
    so I think it could be worthwhile to consolidate all of those cases with a function. Saying that, I completely understand if you don't want to mess with this code further, or think there it is more clear so as to outweigh the benefit of reducing the duplication.

davidhassell:
Thanks for looking at this so carefully, @sadielbartholomew. 

1) That's interesting. Those methods were always there, but further down the list. E.g. `__deepcopy__` came in the `Methods inherited from cfdm.core.abstract.container.Container:` section. More metaclass magic. I personally am not too concerned about this, as some dunderscore methods already get headline billing by virtue of them not being inherited, such as `Field.__getitem__`.

2) I thought the same, and even coded that up to some extent at some point. What made me change my mind was that any metaclass methods would pollute the methods of pretty much all cfdm classes. The methods that are there (e.g. `RewriteDocstringMeta._docstring_special_substitutions` are methods that have value to be called by a user, particularly one who is using the functionality in a child package (like cf-python). That's also why I didn't put them in as functions outside of the class: it complicates the customization of the process should someone want to subclass the metaclass. Happy to revisit thinking about this approach if you like.

sadielbartholomew:
Thank you @davidhassell for the clarifications. In short, that all makes sense and is sensible. I'm just doing a little more testing but I should be happy to approve soon.

> That's interesting. Those methods were always there, but further down the list. E.g. __deepcopy__ came in the Methods inherited from cfdm.core.abstract.container.Container: section. More metaclass magic. I personally am not too concerned about this, as some dunderscore methods already get headline billing by virtue of them not being inherited, such as Field.__getitem__.

Just to clarify I was thinking in particular about the ones relating to the docstring re-writing, e.g. `__docstring_package_depth__`, rather than a generic dunderscore methods. But if you aren't concerned about them being there, that's fine by me. It seems like there is an awful lot documented on the class docstring anyway, such that any users would have to be extremely keen to read through the entire class docstring for bedtime reading or otherwise!

> I thought the same, and even coded that up to some extent at some point. What made me change my mind was that any metaclass methods would pollute the methods of pretty much all cfdm classes. The methods that are there (e.g. RewriteDocstringMeta._docstring_special_substitutions are methods that have value to be called by a user, particularly one who is using the functionality in a child package (like cf-python). That's also why I didn't put them in as functions outside of the class: it complicates the customization of the process should someone want to subclass the metaclass. Happy to revisit thinking about this approach if you like.

No need, those are very strong justifications not to use such a function. I didn't think about or realise those considerations and with them in mind I think your current approach is most sensible. :+1:

sadielbartholomew (approved):
Fantastic! No more mass copy and pasting of duplicated description :tada: And I do agree it is wise to get rid of all those ``.rst.txt`` files, given the sheer amount of them and the collective line count (can't imagine what the equivalent counts were for cf-python!).

FYI as well as an eyeball of the (non-`.rst.txt` removal) changes I checked & observed that:
* inspection with the Python help function gave the formatted docstrings for class, method, function etc.;
* the documentation built with API reference appearing just as before;
* the `test_docstring` test passed locally & is sound (I noted that the method which checks that no `{{` pattern remains in the docstring will not work if we end up using certain ASCII art in future, but I think we can hold back in that respect :wink:).

sadielbartholomew:
(Oh, sorry I assumed you are happy for me to just merge this with approval. If not, feel free to revert as I am not committing anything to `cfdm` at least until tomorrow.)

davidhassell:
All good, thanks for the review, @sadielbartholomew

Issue #53, opened 2020-06-19
Further documentation on creating fields and writing fields to file


sadielbartholomew:
As stated in an email thread:

> An overview of creation and writing near the top of the tutorial (where there is already a read overview) would be beneficial.

This is in response to a(n expert) user request:

> a couple of tutorials on "how to create CF compliant data" would be really useful. We could tie them to useful examples, like: creating a multimodel ensemble dataset and "converting a grib file to cf compliant netcdf"

kurtsansom:
This would be a great help!

Lets say I already have a compliant file, can I extract the schema in a sensible in order to populate a new file with similar but potentially different data. An example like this would be really helpful.

Is there a way to create a "skeleton" of the file and then populate it with data?

sadielbartholomew:
Hi @kurtsansom, thanks for registering that you would also find such further documentation useful. It helps us to know multiple people would benefit from some improvement, as in this case, so we can make it a higher priority.

Your particular questions are interesting because (if I understand your requests correctly) both are indeed possible and also documented quite well, just not with the cfdm library directly, rather in [cf-python](https://ncas-cms.github.io/cf-python/), our data analysis library which extends cfdm in functionality and scope. So, if you are willing to use cf-python instead of cfdm standalone (see [here](https://ncas-cms.github.io/cf-python/installation.html) for the installation guide), I can point you directly to documentation which covers in detail how to do what you wish to, programmatically:

> Lets say I already have a compliant file, can I extract the schema in a sensible in order to populate a new file with similar but potentially different data.
>
> Is there a way to create a "skeleton" of the file and then populate it with data?

Yes, this can be achieved by two different approaches in cf-python:
1. reading in the compliant file (with [`cf.read`](https://ncas-cms.github.io/cf-python/function/cf.read.html?highlight=read#cf.read)) to produce a corresponding list of fields;
2. then either:
    * overwriting the data of these read-in fields (or copies of them if preferred) using the [`set_data` methods on the metadata constructs or entire field](https://ncas-cms.github.io/cf-python/search.html?q=set_data); or
    * running the `creation_commands` method on each read-in field to produce the exact sequence of calls required to generate the fields from scratch (see [this sub-section with an example](https://ncas-cms.github.io/cf-python/tutorial.html?highlight=creation#command-modification)), then running those but with the arguments to the `set_data` call changed as applicable to set different data on the fields;
[...]

PR #82, opened 2020-09-23
Creation commands

davidhassell:
Transfer `creation_commands` methods from cf-python to cfdm, as mentioned in #53

### DIFF
--- a/cfdm/data/data.py
+++ b/cfdm/data/data.py
@@ -1148,6 +1148,153 @@ class Data(Container,
         '''
         return super().copy(array=array)
 
+    def creation_commands(self, name='data', namespace=None, indent=0,
+                          string=True):
+        '''Return the commands that would create the data object.
+
+    .. versionadded:: (cfdm) 1.8.7.0
+
+    :Parameters:
+
+        name: `str` or `None`, optional
+            Set the variable name of `Data` object that the commands
+            create.
+
+        {{namespace: `str`, optional}}
+
+        {{indent: `int`, optional}}
+
+        {{string: `bool`, optional}}
+
+    :Returns:
+
+        {{returns creation_commands}}
+
+    **Examples:**
+
+    >>> d = {{package}}.Data([[0.0, 45.0], [45.0, 90.0]],
+    ...                      units='degrees_east')
+    >>> print(d.creation_commands())
+    data = cfdm.Data([[0.0, 45.0], [45.0, 90.0]], units='degrees_east', dtype='f8')
+
+    >>> d = {{package}}.Data(['N/A', 'beta', 'gamma', 'delta'],
+    ...                      mask = [1, 0, 0, 0])
+    >>> d.creation_commands(name='d', namespace='', string=False)
+    ["d = Data(['', 'beta', 'gamma', 'delta'], dtype='U5', mask=Data([True, False, False, False], dtype='b1'))"]
+
+        '''
+        namespace0 = namespace
+        if namespace is None:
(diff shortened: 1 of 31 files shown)


davidhassell:
(I know that there are still a few TODOs in the docstrings ...)

sadielbartholomew:
About to add my review. Just going to open-close to trigger the CI.

cfdm/test/test_Field.py
```diff
@@ -609,6 +609,34 @@ def test_Field_compress_uncompress(self):
...
+            for indent in (0, 4):
+                for s in (False, True):
+                    for ns in ('cfdm', ''):
+                        _ = f.creation_commands(representative_data=rd,
+                                                indent=indent,
+                                                namespace=ns,
+                                                string=s)
+        # --- End: for
+
+        for i in range(7):
```
sadielbartholomew, line 628: Should probably tidy this a little for readability if nothing else, by moving this `for i in` loop with the example field definition & inner `creation_commands` call under the four nested four loops above, to avoid repeating them.

sadielbartholomew (commented):
Two minor comments, but otherwise all is good. I've tested this end-to-end with cf-python by checking out also the `creation-commands` twin branch, as well as the cfdm standalone test suite, and the test suites both pass locally. I also tried out the new tutorial & API reference examples as a script and they work as they should.

I can't see any issue in the transfer of the relevant documentation, except for one word which should be amended (see sub-comment).

cfdm/data/data.py
```diff
@@ -1148,6 +1148,157 @@ def copy(self, array=True):
...
+        {{returns creation_commands}}
+
+    **Examples:**
+
+    >>> d = {{package}}.Data([[0.0, 45.0], [45.0, 90.0]],
+    ...                      units='degrees_east')
+    >>> print(d.creation_commands())
+    data = cfdm.Data([[0.0, 45.0], [45.0, 90.0]], units='degrees_east', dtype='f8')
+
+    >>> d = {{package}}.Data(['N/A', 'beta', 'gamma', 'delta'],
```
sadielbartholomew, line 1180: Possibly I am missing something obvious, but is there a reason for the 'N/A' here? I;m guessing it is just to indicate that it will be masked so the value is irrelevant? Might be worth commenting as such for clarity as I was a little confused by it at first (could just be me, though).

davidhassell:
Thanks, @sadielbartholomew for the careful review. Does the latest commit sort all of the points you highlighted? If so, please merge! otherwise, let me know ...

cfdm/data/data.py
davidhassell: That's right. See https://github.com/NCAS-CMS/cfdm/pull/82/commits/951ba8bd89d5296db3283e195bae78e7a44148bc

sadielbartholomew: A simple solution, thanks.

sadielbartholomew:
Feedback has all been addressed, thanks @davidhassell.

I have just run the local test suites (cfdm & e2e cf-python) as a final check. Should strictly run the CI but I saw that the PEP8 test is failing the cfdm suite on a small number of cases, so that will fail or cancel the CI jobs anyway.  I can do a commit after this to fix those PEP8 code violations.

This good to merge.

davidhassell:
Thank you.

sadielbartholomew:
Oh, please ignore my comment above about PEP8 failures, I have just seen that they were all arising from a temporary script I specifically made to test which was this living under the repo. So the CI should pass - we can check from the result produced by my merge commit. :+1:

PR #79, opened 2020-09-18
Test update to GitHub Actions (Codecov coverage reports)

sadielbartholomew:
Updates to our workflow to run the test suite, aiming to upload a coverage report to Codecov for two jobs, one each for the latest distros for Ubuntu and Mac using Python 3.7 as a representative Python version (I think it would be superfluous & a waste of Actions jobs to upload further coverage reports, since each job should give a very similar if not identical figure).

Testing in this PR to try to get the two coverage reports to show as a pair of 'checks' tied to the PR. Once I get that working, I can merge the PR.

PR #84, opened 2020-09-25
Customise Codecov configuration

sadielbartholomew:
The purpose of this PR is two-fold:

* tweak the default Codecov config to pass or fail he Github Actions / CI steps to be more appropriate to our current coverage levels and general dev. workflow;
* check that, now at least one Codecov report has been generated on cfdm for the first time, a 'pull request base' is detected and used as a reference for calculating coverage changes (compare to [the initial report](https://github.com/NCAS-CMS/cfdm/pull/79#issuecomment-698808883) where no base was found).

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,35 @@
+# This config overrides the default Codecov config, as outlined at:
+# https://docs.codecov.io/docs/codecov-yaml
+
+coverage:
+  round: up
+  range: "70...100"
+  precision: 2
+
+  status:
+    project:
+      default:
+        target: auto
+        threshold: "0%"
+        base: auto
+        branches: 
+          - master
+        if_no_uploads: success
+        if_not_found: success  
+        if_ci_failed: error
+        only_pulls: true
+    patch:
+      default:
+        target: "90%"
+        threshold: "10%"
+        base: auto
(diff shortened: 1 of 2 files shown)


sadielbartholomew:
The macOS jobs seem to consistently all run a lot slower than the Ubuntu jobs. Not sure why that is but it is something we should keep an eye on in case it is due to some performance issue for Mac distros.

sadielbartholomew:
All checks passed appropriately first time run, but I just edited the workflow so should re-trigger to test I didn't break anything...

codecov.yml
```diff
@@ -0,0 +1,35 @@
...
+coverage:
+  round: up
+  range: "70...100"
+  precision: 2
+
+  status:
+    project:
+      default:
+        target: auto
+        threshold: "0%"
```
sadielbartholomew, line 13: We might want to tweak these targets & thresholds (see also below for the `patch` key) but these seem sensible for now.

sadielbartholomew:
All good and the Codecov report looks legit, albeit rather boring as I didn't change anything that in the codebase that gets tested under coverage (the one line difference is due to non-determinism in the test suite, which is a different matter we should look at at some point).

PR #101, opened 2020-11-17
Docs: add spelling extension and use to fix all typos

sadielbartholomew:
As per the description in the equivalent PR for cf-python, NCAS-CMS/cf-python#153.


sadielbartholomew:
It looks like there are some PEP8 errors to fix, let me sort that now...

sadielbartholomew:
Retriggering the CI with a close-open...

sadielbartholomew:
All checks pass, and there is no change to the test coverage since these changes are to the documentation infrastructure, content, and docstrings, which are currently not tested (though we plan to add tests for these soon including of the build process and of the rendered content e.g. that the code snippets are all working, via the doctest extension).

So this is good to merge.

PR #103, opened 2020-11-19
changes for inheriting packages

davidhassell:
@sadielbartholomew A few minor changes here, that might not even be necessary, but I think they do no harm for now. This has highlighted a bug in that cf-python was updating  the cfdm CONSTANTS dictionary, which is not so good! The forthcoming cf-python PR will fix that issue, but not in a particularly spectacular way, i.e. it's a bodge that is OK for now but we should revisit it later.

### DIFF
--- a/cfdm/functions.py
+++ b/cfdm/functions.py
@@ -139,10 +139,14 @@ def configuration(atol=None, rtol=None, log_level=None):
 
     '''
     return _configuration(
-        new_atol=atol, new_rtol=rtol, new_log_level=log_level)
+        Configuration,
+        new_atol=atol,
+        new_rtol=rtol,
+        new_log_level=log_level
+    )
 
 
-def _configuration(**kwargs):
+def _configuration(_Configuration, **kwargs):
     '''Internal helper function to provide the logic for `cfdm.configuration`.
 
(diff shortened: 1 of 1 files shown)


sadielbartholomew:
Good spot if there was a bug. This all relates to #85, the thrust of which is that we hadn't yet investigated and documented how the constants might interface from cfdm to cf-python.

The `log_level` constant may need special attention, because based on the setup of the logging, a change of the constant in cf-python might not (I can't remember) pass through to amend the log level at which messages emerge in cfdm. We may want to allow the levels of the logging on both projects to vary independently, in which case we should have separate constants in that case, but I think we agreed the remaining constants can be shared? Or would it be best (it would certainly be simpler) to share all of them, assuming independently-varying project logging levels isn't important or useful?

davidhassell:
You're right - I'd forgotten that we agreed the remaining constants can be shared, but I realised that it wasn't playing nicely with the new formulation. I can't remember quite how - but the `atol` constant was not getting reset I after a `with` block because of it. I'll check what I was doing ... 

In https://github.com/NCAS-CMS/cf-python/pull/155, and in this PR, I did indeed put in some code for `log_level`

PR #104, opened 2020-11-20
Constants inheritance proposal #1

davidhassell:
Hi @sadielbartholomew here are some thoughts for refactoring the constants access situation so that we can most effortlessly use it in cf-python. The user-facing API is unchanged. What do you think?

The on-line docs are unchanged. There is a small change to the `help(cfdm.rtol)` docs, but tolerable, I think.

cf-python would use it as follows:

```python
# in cf/functions.py

class rtol(cfdm.rtol):
    _CONSTANTS = CONSTANTS
    _Constant = Constant


rtol.__doc__ = cfdm.rtol.__doc__.replace('cfdm.', 'cf.')
```

So cf-python need not know anything about the functionality nor API nor docs from cfdm.

This is just one idea of many, I'm sure.


sadielbartholomew:
Hi @davidhassell, I think that is an excellent solution: minimal code required to implement the inheritance, as per your example snippet, and it can keep the logic loosely coupled:

> So cf-python need not know anything about the functionality nor API nor docs from cfdm.

Good thinking.

I was trying to work out whether the `__doc__` replacement would work in light of us running the docstring rewriting logic on all of the docstrings (even if they don't themselves have anything that would be replaced by it), but I think it shouldn't be a problem since the docstring rewriting is done via metaclass which should happen before the replacement happens. I could be wrong though, have you checked it works in this respect (if it happened the other way round the replacement might be futile, if that makes sense)?

davidhassell:
Thanks. Good point on doc string rewriting. Currently it is not in place for these new classes, but I reckon if we add simply the metaclass to `ConstantAccess` the magic will just happen ... I'll give it a go. The doc string _overwriting_ is somewhat clunky, so it's nice to get rid of it.

sadielbartholomew:
> Currently it is not in place for these new classes

Oh, I didn't consider this, so I guess a simple option is just to not apply it to those classes, if it does end up causing issue!

On the topic the `atol` and `rtol` docstrings are pretty similar so to enforce consistency they may be a good candidate for one or more common text snippets to process in, though being only two cases it might be overkill.

All good. If you would like review once this is ready just ping me. Thanks.

davidhassell:
Will do. I've just remembered, after a bit of tinkering, that the docstring rewriting only happens on methods of the class, rather than the class itself.

It may/may not be trivial to extend it to the `__doc__` of the class. For methods, the original method gets copied, the copy's doc string is rewritten and then the new method replaces the old one. This is so that docstring changes do not affect other classes that inherit the same method.

Can we somehow copy the class itself in the same way. I have no idea, but it'll fun(?) finding out.

If this is not easy to fix, I suggest that we implement it as it stands and dwell on the more elegant solution at our leisure.

davidhassell:
(I should say that adding the metaclass absolutely did the trick for _method_ docstring rewriting on the `rtol` class)

sadielbartholomew:
> Can we somehow copy the class itself in the same way. I have no idea, but it'll fun(?) finding out.

I think that should be do-able via an approach not dissimilar to one used for copying the methods. But I could be wrong. Would not some investigation, Good luck, if you want to try it.

> If this is not easy to fix, I suggest that we implement it as it stands and dwell on the more elegant solution at our leisure.

Sounds like a good plan to me. I think the class copying, as above, constitutes another issue entirely, so perhaps we can throw an issue up for it to tackle at a good time.

davidhassell:
Class docstring rewriting is a reality . I think it's all there, now, for now.

cfdm/core/meta/docstringrewrite.py
```diff
@@ -517,21 +577,24 @@ class that inherits such such a method will need to exclude it,
 
     @classmethod
     def _docstring_update(cls, package_name, class_name, f,
-                          method_name, config):
+                          method_name, config, class__doc__=None):
```
sadielbartholomew, line 580: I feel `class__doc__` could be a slightly confusing name for an argument. Though it is very clear what it is referring to, at a glance I personally would see the double underscores and without looking carefully think it was `class.__doc__` when used internally. So, it's not crucial, but if you agreed for our future benefit perhaps renaming to `class_docstring` or similar would be better.

cfdm/functions.py
```diff
@@ -17,9 +17,25 @@
...
+from .core.docstring import (_docstring_substitution_definitions
+                             as _core_docstring_substitution_definitions)
+
+from .docstring import _docstring_substitution_definitions
+
 from .constants import CONSTANTS, ValidLogLevels
 
 
+# --------------------------------------------------------------------
+# Merge core and non-core docstring sustituion dictionaries
```
sadielbartholomew, line 31:
```suggestion
# Merge core and non-core docstring substitution dictionaries
```

cfdm/functions.py
```diff
@@ -17,9 +17,25 @@
...
+# --------------------------------------------------------------------
+_subs = _docstring_substitution_definitions.copy()
+_subs.update(_core_docstring_substitution_definitions)
+_docstring_substitution_definitions = _subs
+del _subs
```
sadielbartholomew, line 36: It might be worth adding a brief comment explaining why `_docstring_substitution_definitions` can't be updated directly, in case this leaves anyone scratching their heads.

cfdm/functions.py
```diff
@@ -998,38 +747,86 @@ def copy(self):
...
-
     def __new__(cls, *args, **kwargs):
+        '''Must override this method in subclasses.
+
+        '''
         instance = super().__new__(cls)
         instance._func = configuration
         return instance
 
+    def __docstring_substitutions__(self):
```
sadielbartholomew, line 793: When checking the docs built off this branch, I've noticed any dunderscore methods aren't currently processed and included in the API reference. If we want to include them, we'll need to tell Sphinx we want them. This worked for me (I'm using `v3.2.1`):

```diff
diff --git a/docs/source/conf.py b/docs/source/conf.py
index 041303e43..fba9321bb 100755
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -130,6 +130,7 @@ autodoc_member_order = 'groupwise'
 autodoc_default_options = {
     'members': True,
     'inherited-members': True,
+    'special-members': True,
     'show-inheritance': True,
 }
```

sadielbartholomew (approved):
I have a small number of minor comments, but overall this is great and it works perfectly (as far as I can reasonably determine) to retain the behaviour and documentation of the `cfdm` constants and functions for getting & setting them, while changing the underlying implementation so maintenance is easier and inheritance by cf-python is more robust. :tada:

cfdm/functions.py
davidhassell: Good point. As of Python 3.9 PEP 584  this is much more elegant

cfdm/functions.py
davidhassell: I'll need to think about this one, just because I can't remember what the implications are for the appearance of classes in general ....

sadielbartholomew: Sounds wise :) Please take your time deciding, after all it is easy to add this or similar in at a later date.

davidhassell:
HI Sadie - I think it's OK to merge now?

cfdm/functions.py
davidhassell: Hi Sadie, when I set `special-members` at v2.3.1 nothing happens, possibly because we set `no-members` on all of our `autolcass` directives? I tend to only document the dunderscore methods that have been overridden, but that's not  consistently applied, either. I'll try at 3.2.1 and see what happens ...

(pause of a few minutes)

... Wow. 3.2.1 destroyed the docs, with or without `special-members` in different ways. It's actually trashed all of the files  in `docs/source/class`! (not problem in getting them back, though) 

Can we raise this as a separate issue, that also will address the "right" version of sphinx to use? I'm sure it should be newer than my 2.3.1, and I may just need to update something else in my environment (I hope)

cfdm/functions.py
```diff
@@ -1062,3 +848,426 @@ def __repr__(self):
...
+    Subclasses must implement or inherit a method called `_parse` as
+    follows:
+
+       def _parse(cls, arg):
+          """Parse a new constant value.
+
+       :Parameter:
+
+            cls:
+                This class.#
```
sadielbartholomew, line 890:
```suggestion
                This class.
```

sadielbartholomew:
> HI Sadie - I think it's OK to merge now?

I found a hash symbol lurking, as in my above comment, but otherwise the minor feedback is all addressed and everything is shipshape, so please merge! I'll start reviewing the cf-python follow-on PR this afternoon.

davidhassell:
Great - thanks

PR #105, opened 2020-11-23
Decorator to implement display argument print or return

### LINKED ISSUE #96: Decorator to administer display keyword argument
This is very low on the prioritisation list, but I've noticed there are several cases of logic to either return or print a constructed string keyword parameter depending on the value of a Boolean parameter display, essentially (in the most common case) the following:

```python
def <method name>(<args, other kwargs>, display=True):
    <...
    construct 'string' var
    ....>

    string = '\n'.join(string)

    if display:
        print(string)
    else:
        return string
```

and this would be a great candidate for logic to apply via a decorator. It should be fairly to implement such a decorator, too.

sadielbartholomew:
Closes #96, including adding a test for this new decorator which I have called `_display_or_return`, where I have used `@unittest.mock.patch('builtins.print')` to explicitly check that any case that should print something does indeed print it, as well as checking for (lack of) output.

### DIFF
--- a/cfdm/decorators.py
+++ b/cfdm/decorators.py
@@ -22,7 +22,7 @@ INPLACE_ENABLED_PLACEHOLDER = '_to_assign'
 def _inplace_enabled(operation_method=None, *, default=False):
     '''A decorator enabling operations to be applied in-place.
 
-    If the decorated method has keyword argument *inplace* being equal
+    If the decorated method has keyword argument ``inplace`` being equal
     to True, the function will be performed on `self` and return None,
     otherwise it will operate on a copy of ``self`` and return the
     processed copy.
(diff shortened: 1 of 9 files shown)


sadielbartholomew:
Forgot to write the docstring of the new decorator, so have pushed a commit to do that. The CI tests all passed prior to this new commit, but I'll re-trigger them to re-run the checks just to be safe.

sadielbartholomew:
All good, therefore merging.

cfdm/decorators.py
```diff
@@ -22,7 +22,7 @@
 def _inplace_enabled(operation_method=None, *, default=False):
     '''A decorator enabling operations to be applied in-place.
 
-    If the decorated method has keyword argument *inplace* being equal
+    If the decorated method has keyword argument ``inplace`` being equal
```
davidhassell, line 25: I think we need a policy on **\*** vs. **\`\`** vs. **\`**!

The unwritten policy that I've been following is to use **\*** for parameters in the abstract (rendering as italics), and **\`\`** for anything that might by typed or that appears on the screen, and  **\`** being reserved for functions, classes, methods, and attributes (rendering as (usually) clickable links in the online docs).

By this reckoning, inplace would be `*inplace*`

This is a framework that has evolved organically over many years, largely driven by the tech in use at any one time to render the docs, so very happy to review this never-reviewed procedure!

Perhaps it's best to use `*inplace*` for now? Then if we were to decide on a new approach, it would be easier to find and change.

sadielbartholomew: Ah, sorry I did not realise/recall this was a somewhat controversial change, I made it in the spur of the moment because it the italicised keyword didn't seem right at an eyeball! But now you point it out I have a vague recollection of you mentioning this early in the year, so I think there was a policy and just I overlooked it :grimacing:

I'm happy to continue with the "unwritten policy", I'll pop in a commit now to fix this case I defaced and also the equivalent case of the new display argument.

(Also, sorry if you wanted to do a review. I didn't assign you or ask if you wanted to review this because this PR was pretty straightforward.)

sadielbartholomew: Sorted here: 548584e38ea189cd5cbc81dda0179c0fed2f3a7e. I have made a mental note to continue with the asterisk/backtick policy you outlined above (assuming you aren't desperate to change it).

davidhassell: Hi Sadie, that's all good. I'm not at desparate to change it - if/when the current style isn't doing what we want any more is probably the time to review it.

PR #107, opened 2020-11-26
Fix for resetting constants when configuration call fails

davidhassell:
Towards addressing https://github.com/NCAS-CMS/cf-python/pull/155#pullrequestreview-538856546


sadielbartholomew (approved):
Tested as making the required fix for cfdm. Thanks.

PR #108, opened 2020-12-07
Allow functions to be decorated with via_verbosity

sadielbartholomew:
The `_manage_log_level_via_verbosity` decorator was only applicable and functioning for use on methods, not for functions not bound to classes, but there are a number of cases of the latter in both `cfdm` and `cf` where we want to apply it. This PR:

* confirms the above via extending the decorator unit test so there is failure when it is applied to a function; &
* generalises the decorator so it can be used in any case, fixing the new test failures (this was easy as `self` can be included in the `*args` already included in the calls, so I just had to remove the explicit reference to `self`).

### DIFF
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -103,7 +103,7 @@ jobs:
       # would have to run anyway, it is simplest to add it in this flow.
       # Also, it means code coverage is only generated if the test suite is
       # passing at least for that job, avoiding useless coverage reports.
-      uses: codecov/codecov-action@v1.0.13
+      uses: codecov/codecov-action@v1
       if: |
         matrix.os == 'ubuntu-latest' && matrix.python-version == 3.7
       with:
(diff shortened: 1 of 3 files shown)


sadielbartholomew:
Re-triggering the CI because one of the tests failed with some Codecov upload issue:

```
Error: Codecov failed with the following error: Hostname/IP does not match certificate's altnames: Host: codecov.io. is not in the cert's altnames: DNS:prod01-web.codecov.io
```

sadielbartholomew:
The same error emerged again, so I've bumped the version (by generalising it) of the external Action for Codecov which hopefully will resolve that. Re-triggering again now.

sadielbartholomew:
Nope, that hasn't fixed it which is frustrating since the Codecov Action has been working fine for months. I am going to assume it is an intermittent issue and re-try the CI tomorrow.

sadielbartholomew:
... our Codecov page is inaccessible which backs up my theory above.

sadielbartholomew:
Codecov page is back, suggesting the intermittent are issues over, so retrying...

sadielbartholomew:
Got there eventually! All checks have passed so this is now good to merge.

PR #112, opened 2021-01-14
Back in black: format code and set up pre-commit hooks

sadielbartholomew:
Equivalent to sibling PR NCAS-CMS/cfunits#21; see the description there which is precisely applicable here.


sadielbartholomew:
Deleting the Codecov comment as it is not relevant here since no functional changes have been made, rather purely formatting and/or style.

Merging since follow-on work can be completed in further PR(s). I have already ran into several sets of merge conflicts by keeping this open. Follow-on work is to:

* adapt the docstring rewriting functionality as necessary to the extra level of indentation `black` applies to ensure it renders the end docstrings as before;
* add further validation for docstring line length (PEP 257) and format via the `docformatter` library as a checker (on Actions) and a pre-commit hook enforcer;
* change docstrings to start on the line after the opening quotes, to make lines start at a consistent level and help enforce docstring line length with the above.

PR #115, opened 2021-01-20
Enable faulthandler for all test modules

sadielbartholomew:
See description to NCAS-CMS/cf-python#179 which this PR is equivalent to, but for cfdm.


sadielbartholomew:
Currently no seg faulting or timing out on the GitHub Actions cfdm test suite as expected, though I do personally see seg faults locally which are under investigation. No change in test coverage as nothing functional or test-based was added other than the switching on of `faulthandler`. So good to merge,

PR #117, opened 2021-01-29
_custom documentation

davidhassell:

PR #118, opened 2021-01-29
_custom extension docs

### LINKED ISSUE #116: Container copy method implies deep copy behaviour
Initial work towards doctesting has implied, and after further investigation I can confirm, that the `copy` method for the ABC `Container` (i.e. `cfdm.core.Container.copy`) that is [documented as being a deep copying operation](https://ncas-cms.github.io/cfdm/method/cfdm.core.Container.copy.html#cfdm.core.Container.copy) is in fact only displaying the behaviour of a shallow copy.

For example, note how the setting of a `_custom` dict component of `g` is also reflected in `f` when it is an item within a container, appearing to be a reference rather than a copy of that item, but not reflected in `f`f when a simple object:

```python
>>> # Setup
>>> import cfdm
>>> f = cfdm.core.abstract.container.Container()
>>> f._custom
{}
>>> f._custom['feature'] = ['f']
>>> f._custom
{'feature': ['f']}

# Apply the copy, expecting it to be deep
>>> g = f.copy()
>>> g._custom['feature'][0] = 'g'
>>> g._custom
{'feature': ['g']}

# ...but note how the change is also reflected in f:
>>> f._custom
{'feature': ['g']}

# ...though changing the top-level value for g does not influence f:
>>> g._custom['feature'] = 'gee whiz'
>>> g._custom
{'feature': 'gee whiz'}
>>> f._custom
{'feature': ['g']}
```

### Environment

```ipython
Platform: Linux-4.15.0-54-generic-x86_64-with-glibc2.10
HDF5 library: 1.10.6
netcdf library: 4.7.4
Python: 3.8.5
netCDF4: 1.5.4
numpy: 1.19.4
cfdm.core: 1.8.8.1
cftime: 1.3.0
netcdf_flattener: 1.2.0
cfdm: 1.8.8.1
```

davidhassell:
with #117, fixes #116

Description in extensions.rst of how the `_custom` dictionary works

### DIFF
--- a/docs/source/extensions.rst
+++ b/docs/source/extensions.rst
@@ -10,14 +10,13 @@
 
 Version |release| for version |version| of the CF conventions.
 
+.. contents::
+   :local:
+   :backlinks: entry
 
 The cfdm package has been designed to be subclassed, so that the
 creation of a new implementation of the CF data model, based on cfdm,
 is straight forward. For example:
-
-
-.. Creating a new implementation of the CF data model by subclassing
-   cfdm classes is straight forward. For example:
 
 .. code-block:: python
    :caption: *Create a new implementation with a new field construct
@@ -115,8 +114,7 @@ in overridden methods.
    
    >>> class my_Field_2(cfdm.Field):
    ...    def my_coordinates(self):
-   ...        '''Get coordinate constructs with a different API
-   ...        '''
+   ...        """Get coordinate constructs with a different API."""
    ...        c = self.coordinates
    ...        if not c:
    ...            return {}
@@ -124,9 +122,9 @@ in overridden methods.
    ... 
    >>> class my_CFDMImplementation(cfdm.CFDMImplementation):
    ...    def get_coordinates(self, field):
-   ...        '''Get coordinate constructs from a my_Field_2 instance,
-   ...    using its different API.
-   ...        '''
+   ...        """Get coordinate constructs from a my_Field_2 instance,
+   ...        using its different API.
+   ...        """
    ...        return field.my_coordinates()
    ... 
    >>> my_implementation_2 = my_CFDMImplementation(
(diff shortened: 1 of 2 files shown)


docs/source/extensions.rst
```diff
@@ -216,8 +216,45 @@ elements in a dataset.
...
+private quantities which are tightly controlled); or else include
+extra code that does deep copy such mutables when any deep copy (or
+equivalent) operation is called. The latter should be implemented in
+the subclasses `__init__` method, similarly to this:
 
+.. code-block:: python
+   :caption: *Ensure that the _custom dictionary 'x' value is deep
+             copied when a deep copy of an instance is requested.*
+	 
+   >>> class my_Field_3(cfdm.Field):
```
sadielbartholomew, line 245: In the case of this code example, and perhaps some of the others on the 'extension' page where they are no outputs to show, could it be clearer to show it as a pure Python block rather than in iPython with the prompts and need for continuation characters afterwards?

sadielbartholomew (approved):
The description is nice and clear so I'm very happy with this (and the one CI failure is for code coverage which is irrelevant as this is docs-related), thanks, though I've noted a few minor comments.

docs/source/extensions.rst
davidhassell: Good idea - I'll sort this out on Monday ...

sadielbartholomew: Please do, it is solidly into weekend time now :grinning: Have a good one.

davidhassell:
@sadielbartholomew How's that?

sadielbartholomew (approved):
> How's that?

Perfect! Minor feedback addressed and the built extensions HTML Page is all formatted fine. The final description is very clear. Thanks.

PR #120, opened 2021-02-05
Docformatter + pydocstyle: hooks and standardisation

sadielbartholomew:
See description to NCAS-CMS/cfunits#22 which this PR is equivalent to for cfdm.


sadielbartholomew:
I'll address the minor merge conflicts and merge next week.

I'm also going to raise an issue on the pydocstyle Issue Tracker because something about one of our docstrings is confusing pydocstyle (see the note in the commit message for 61a51d5). I think I know what it is, but there is seemingly no way to address the corresponding docstring format errors on our end, and it would be nice not to always have to ignore the pydocstyle hook for commits touching the `functions` module for that reason.

sadielbartholomew:
Closing and re-opening to trigger the CI for a final check, though only the docstrings and some config has changed and I haven't added any docstring checking to an Actions workflow yet, so it will not check the same aspects as the pre-commit hooks yet...

sadielbartholomew:
All good to merge.

Issue #128, opened 2021-03-26
Mapping of metadata-only CDL to field constructs


sadielbartholomew:
NCAS-CMS/cf-python#197 has highlighted that `cfdm` as-is is not able to interpret as definite field(s) some cases of CDL inputs which provide only schema and/or coordinate information, e.g. as produced from a `ncdump -h` or `ncdump -c`. We realised it may not in fact be possible given the nature of the CF data model to unambiguously map CDL with metadata but no data arrays provided onto fields, though we aren't certain.

The direct problem resulting from this is that such missing data is not accounted for so errors may emerge when such CDLis read-in.

### Long-term solution

Ultimately we should:

1. determine whether header- and coordinate-based CDL can be mapped conclusively to field constructs,
2.
    a) if so, adapt the logic so that missing data is dealt with appropriately and does not emerge downstream as errors for reading of such CDL, both here in standalone cfdm or indeed in dependencies like cf-python, like as for NCAS-CMS/cf-python#197;
    b) if *not* so, decide on the best approach for a result of reading in such CDL, whether that is to provide some sensible "best guess" field encapsulating the known information, or to tell users via appropriate error message that the CDL can't be read-in as fields for these reasons.

### Short-term fix

For the forthcoming release, to address NCAS-CMS/cf-python#197, I am catching the `MaskError`s to raise them as a user-friendly message stating that the CDL metadata is insufficient for conversion to field constructs, i.e. assuming case (b) and raising a `ValueError`. This is sufficient for the release but should be re-evaluated in the longer term.

Even if we end up going with this approach after the review, I would like to create a custom error class to raise for related errors, rather than using a Python built-in Exception.

PR #129, opened 2021-03-26
Short-term fix for errors in reading of metadata-only CDL

sadielbartholomew:
See #128 for details.


sadielbartholomew:
Fair point from Codecov, it would certainly be good to have an extra test to cover metadata-only CDL and potential errors arising from reading it in, ensuring they are handled okay. I'll add one to this PR next week, so hold on merging.

sadielbartholomew:
The new commit adds testing to cover the cases of relevance. Opening and closing to re-trigger the CI tests...

sadielbartholomew:
The Codecov 'patch' change is sufficient but not the 'project' change, though having looked through the details I am satisfied the latter case is not an issue in context.

All the other tests pass both on the CI and locally, so this is ready to merge.

Issue #132, opened 2021-04-07
Change constructs access API for consistency


davidhassell:
### Current situation

The `construct` and `construct_key` methods take a single positional `identity` argument (https://github.com/NCAS-CMS/cfdm/blob/v1.8.8.0/cfdm/mixin/constructaccess.py#L301):
```python
    def construct(self, identity, default=ValueError()):
```

### Proposed change

The `construct` and `construct_key` methods take any number of positional arguments, as well as new keyword arguments:
```python
    def construct(self, *identity, default=ValueError(), **filter_kwargs):
```
The `**filter_kwargs` are passed to the new `Constructs.filter` method for further selection  flexibility.

This makes sense, as you can already match on multiple identities by providing a `re.Pattern` object, so it doesn't makes sense to restrict this to, say, `*["latitude", "longitude"]`

### Consequences

The only possibly bad effect that this could have is for the  `default` argument to be consumed by the `*identity` arguments if it is given as a positional parameter. For example, to Specify a default of `None`:

```python
>>> f.construct('altitude', None)  # ok at v1.8.8.0
>>> f.construct('altitude', None)  # NOT ok with proposed change
>>> f.construct('altitude', default=None)  # ok with proposed change and at v1.8.8.0
```

Regardless of this potential for code breaking, I think this change will be beneficial.

PR #145, opened 2021-06-07
Improve handling of constructs by `Field.convert`

sadielbartholomew:
Update the `Field.convert` method to:

* provide a better error message and clarify in the docstring that only metadata construct inputs that *can* have data (i.e. are able to, but if they don't that's okay, opposed to ones that can't ever) can be converted i.e. are valid; and
* apply the new (v1.8.9.0) API for construct selection (see #132) since this method seems to have been missed with that update.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,12 @@
+version 1.8.?.?
+---------------
+----
+
+**2021-??-??**
+
+* Construct access API changes from 1.8.9.0 applied to `Field.convert`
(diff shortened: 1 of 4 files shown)


sadielbartholomew:
One of the jobs ran against the previous commit failed on some Codecov API query issue, but most of the other test run jobs that had completed by the time it failed and cancelled the others had passed, plus the linting job passed, so I am satisfied this is good to merge and won't run the jobs again.

PR #146, opened 2021-06-28
Complete docstring TODOs, removing duplication

sadielbartholomew:
Trivial PR to add the collection of daily commits I have made towards addressing the remaining TODOs for populating empty docstrings (normally I would push these individually as I do them, but I got behind early on in the process of doing these, hence a PR to keep the commit history clean).

In the process of doing this I have noticed, and remedied, that:
* a few test methods were, presumably by copy-and-paste error, duplicates of other (see d5563ef);
* some tests classes had duplicate names (see ab38070 and cc0e4e9) and for safety it was bets to rename these distinctly.

### DIFF
--- a/cfdm/test/create_test_files.py
+++ b/cfdm/test/create_test_files.py
@@ -20,7 +20,7 @@ VN = cfdm.CF()
 # DSG files
 # --------------------------------------------------------------------
 def _make_contiguous_file(filename):
-    """TODO DOCS."""
+    """Make a netCDF file with a contiguous ragged array DSG feature."""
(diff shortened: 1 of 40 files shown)


sadielbartholomew:
One test failure causing a test suite run job to fail. I've just run it locally and can see why it is failing. I will push up another commit to fix it.

sadielbartholomew:
Re-running the CI jobs, triggering via open-close...

sadielbartholomew:
Trivial PR so no need for peer review. All checks have passed, so good to merge.

PR #149, opened 2021-07-23
Apply & add hook for isort (systematic import sorting)

sadielbartholomew:
As per title description. No functional changes made, just automatic re-organisation of import statments. See NCAS-CMS/cf-python#63 for context: I am adding PRs now to apply `isort` across cfunits, cfdm and cf-python in turn.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -20,8 +20,8 @@ jobs:
   # *The config. applied here is from our .pre-commit-config.yaml*
   #
   # Note the pre-commit linting includes all of the other linters configured
-  # to run as pre-commit hooks, namely black, flake8, docformatter and
-  # pydocstyle, so there is no need to set these up individually!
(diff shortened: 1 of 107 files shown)


sadielbartholomew:
OK, from the CI results it looks like `black` and `isort` are not quite configured to be compatible, as I thought I had achieved, for some cases relating to mostly line wrapping of long import lists, so I need to fix that. I'll push up another commit to address it shortly.

sadielbartholomew:
> from the CI results it looks like black and isort are not quite configured to be compatible

The issue was just the line length, where I had neglected to consider that we override from the (silly) `black` default  it to be `79` as per PEP8 guidelines. All the hooks pass locally now after I set `isort` to use line length 79 and re-applied `black` and `isort` manually across the codebase.

sadielbartholomew:
All good to merge now, with the one CI job failing being the project coverage which is not relevant here because these are non-functional changes, ultimately just re-organising import statements.

PR #171, opened 2022-01-12
Remove generator.close()

### LINKED ISSUE #170: `filter_by_identity` fails when `identities` returns an `itertools.chain` object
At versions 1.8.9.0 to 1.9.0.1, `filter_by_identity` fails when `identities` returns an `itertools.chain` object
```
cfdm/constructs.py",
line 1980, in _filter_by_identity
     generator.close()
AttributeError: 'itertools.chain' object has no attribute 'close'
```
In this case, the `itertools.chain` object is being passed in by cf-python (https://github.com/NCAS-CMS/cf-python/blob/v3.11.0/cf/mixin/propertiesdata.py#L4808).

I don't think that the generator needs closing here - I can't divine any memory or speed benefit, so I think just deleting that line will suffice. PR to follow.

davidhassell:
Fixes #170

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -6,6 +6,9 @@ Version 1.9.?.?
 * Fixed bug that caused a `cfdm.write` failure when a vertical
   coordinate reference construct has no coordinates
   (https://github.com/NCAS-CMS/cfdm/issues/164)
+* Fixed bug that caused a failure when downstream `identities` methods
+  return an `itertools.chain` object
+  (https://github.com/NCAS-CMS/cfdm/issues/170)
 
 ----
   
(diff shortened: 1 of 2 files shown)


davidhassell:
There's no unit test, as this can't currently happen in pure cfdm - it need to instigated by cf-python. I'll think about putting in a test on the other repo.

sadielbartholomew (approved):
Fixes the issue at hand, and I am confident enough it is safe to remove the line in question, agreeing that the generator `close` is unnecessary in the first place, since the generator in question is not abandoned at any point, nor is it compound (no internal `try`/`except`/`finally` statements or delegation to other generators, etc.).

sadielbartholomew:
We can ignore the Codecov failure as justified by David:

> There's no unit test, as this can't currently happen in pure cfdm - it need to instigated by cf-python. I'll think about putting in a test on the other repo.

so this is now good to merge.

PR #173, opened 2022-02-18
Update deprecated `distutils.version` to `packaging.version`

sadielbartholomew:
Equivalent to NCAS-CMS/cf-python#324 but for cfdm, which also uses the deprecated API calls in question, this time across several modules.


sadielbartholomew:
Sorry @davidhassell I didn't notice there were merge conflicts to resolve for this PR. That is now done.

Please confirm it's OK to add the (essentially necessary) new dependency `packaging` here, with the same context as in https://github.com/NCAS-CMS/cf-python/pull/324#issue-1142096607, then I can merge this and we won't be spammed about the deprecated API use.

davidhassell:
Thanks, Sadie - all OK with me.

davidhassell (approved):

PR #175, opened 2022-02-28
Fix for print when no dimension coordinate data

### LINKED ISSUE #174: Field/domain print fails when dimension corodinate has no data
(First reported over at https://github.com/NCAS-CMS/cf-python/issues/326 - see there for full details)

When a field has a construct which can have data, e.g. a coordinate, but that construct doesn't have data, the `str()` representation hits a `ValueError` regarding the lack of data on that construct, e.g:

```python
>>> import cfdm
>>> f = cfdm.example_field(2)
>>> t = f.construct('time')
>>> t.del_data()
<Data(36): [1959-12-16 12:00:00, ..., 1962-11-16 00:00:00]>
>>> print(f)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-fc0364975534> in <module>
----> 1 print(f)

~/miniconda3.9/lib/python3.9/site-packages/cfdm/field.py in __str__(self)
    239             string.append(f"Field ancils    : {field_ancils}")
    240 
--> 241         string.append(str(self.domain))
    242 
    243         return "\n".join(string)

~/miniconda3.9/lib/python3.9/site-packages/cfdm/domain.py in __str__(self)
    194                 if construct_data_axes[cid] == (axis_cid,):
    195                     name = dim.identity(default=f"key%{0}")
--> 196                     y = f"{name}({dim.get_data().size})"
    197                     if y != axis_names[axis_cid]:
    198                         y = f"{name}({axis_names[axis_cid]})"

~/miniconda3.9/lib/python3.9/site-packages/cfdm/core/abstract/propertiesdata.py in get_data(self, default, _units, _fill_value)
    290                 return
    291 
--> 292             return self._default(
    293                 default, message=f"{self.__class__.__name__} has no data"
    294             )

~/miniconda3.9/lib/python3.9/site-packages/cfdm/core/abstract/container.py in _default(self, default, message)
    145                 default = type(default)(message)
    146 
--> 147             raise default
    148 
    149         return default

ValueError: DimensionCoordinate has no data

>>>
```

Something sensible should be displayed, instead.

davidhassell:
Fixes #174

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,14 @@
+Version 1.9.?.?
+---------------
+
+**2022-0?-??**
+
+* Fixed bug that caused `cf.Domain.__str__` to fail when a dimension
+  coordinate construct does not have data
+  (https://github.com/NCAS-CMS/cfdm/issues/174)
+
+----
+  
 Version 1.9.0.2
 ---------------
 
--- a/cfdm/domain.py
+++ b/cfdm/domain.py
@@ -187,15 +187,18 @@ class Domain(
 
         x = []
         dimension_coordinates = self.dimension_coordinates(todict=True)
-        for axis_cid in sorted(self.domain_axes(todict=True)):
+        for axis_cid, axis in sorted(self.domain_axes(todict=True).items()):
             for cid, dim in dimension_coordinates.items():
                 if construct_data_axes[cid] == (axis_cid,):
                     name = dim.identity(default=f"key%{0}")
-                    y = f"{name}({dim.get_data().size})"
+                    y = f"{name}({axis.get_size()})"
                     if y != axis_names[axis_cid]:
                         y = f"{name}({axis_names[axis_cid]})"
+
                     if dim.has_data():
                         y += f" = {dim.get_data()}"
+                    else:
+                        y += " = "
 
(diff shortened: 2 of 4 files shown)


davidhassell:
(seems like I updated a joss bibtex date - by mistake? I'm not sure. We can let it go through and I'll sort it out with a commit afterwards ...)

sadielbartholomew (approved):
Perfect, thanks. Please merge.

In reviewing, I went to also check that deleting the data from any of the other constructs that take data wouldn't cause the `__str__` output to error, and indeed they don't cause any issues (I checked for an auxiliary coordinate, field ancillary, cell measure and domain ancillary), so I agree with your identification that only the deletion of dimension coordinates' data caused issue (despite my issue write-up implying it could be a more general issue). 

Still, if they aren't covered elsewhere, maybe we could add in some equivalent testing to the small block you added to 'test_Domain' under 'Test when dimension coordinate has no data' to cover those other constructs no having data? I can do that as a follow-up commit or PR, unless you want to add it in here?

davidhassell:
Tanks, Sadie - let's do as you suggest in a follow up PR.

PR #180, opened 2022-03-21
Construct access API

### LINKED ISSUE #179: Add new construct retrieval methods
It ouwld be useful to have the following new methods, which would have the same behaviour and API as their cf-python counterparts:

* `cfdm.Field.auxiliary_coordinate`
* `cfdm.Field.cell_measure`
* `cfdm.Field.cell_method`
* `cfdm.Field.coordinate`
* `cfdm.Field.coordinate_reference`
* `cfdm.Field.dimension_coordinate`
* `cfdm.Field.domain_ancillary`
* `cfdm.Field.domain_axis`
* `cfdm.Field.field_ancillary`
* and corresponding methods on the `cfdm.Domain` class, as appropriate.

(for example, see `cf.Field.dimension_coordinate`: https://github.com/NCAS-CMS/cf-python/blob/v3.12.0/cf/mixin/fielddomain.py#L2061-L2125).

This makes sense from a code point of view following on from the changes implemented in #130 and #132; and makes sense from a user perspective, as these new methods are intuitive to use, and users (well, me, at least!) of cf-python keep trying to use them on cfdm and being frustrated that they're not there.

davidhassell:
Fixes #179


docs/source/tutorial.rst
```diff
@@ -1525,42 +1521,79 @@ construct key, by any of the following techniques:
...
+   >>> key = t.construct('latitude', key=True)
    >>> c = t.constructs.get(key)
    <AuxiliaryCoordinate: latitude(10, 9) degrees_N>
 
-The `~Field.construct` method of the field construct and the
-`~Constructs.value` method of the `Constructs` instance will raise an
-exception of there is not a unique metadata construct to return, but
-this may be replaced with returning a default value or raising a
-customised exception:
+* with the `~Constructs.value` method of a `Constructs` instance, or
```
sadielbartholomew, line 1528: Since when I look at the rendered docs this seems to be the final item in the list:

```suggestion
* with the `~Constructs.value` method of a `Constructs` instance.
```

docs/source/tutorial.rst
```diff
@@ -1525,42 +1521,79 @@ construct key, by any of the following techniques:
...
+             its construct identity, and also its key.*
+	     
+   >>> t.auxiliary_coordinate('latitude')
+   <CF AuxiliaryCoordinate: latitude(10, 9) degrees_N>
+   >>> t.auxiliary_coordinate('latitude', key=True)
+   'auxiliarycoordinate0'
+   >>> t.auxiliary_coordinate('latitude', item=True)
+   ('auxiliarycoordinate0', <CF AuxiliaryCoordinate: latitude(10, 9) degrees_N>)
+
+All of these techniques will raise an exception of there is not a
```
sadielbartholomew, line 1567:
```suggestion
All of these techniques will raise an exception if there is not a
```

sadielbartholomew (approved):
Some minor comments, but overall this is great: the new methods all work as they should, and sufficient testing and detailed documentation has been added accordingly. Please merge when ready.

PR #168, opened 2021-12-21
Subsampled coordinates (1) and compression refactor

davidhassell:
Fixes #167

PR #183, opened 2022-04-27
Documentation: avoid use of double negatives for clarity

sadielbartholomew:
A documentation improvement discussed previously: see https://github.com/NCAS-CMS/cfdm/pull/168#discussion_r816857213 for context. This PR aims to find all uses of double negatives (in practice these are in keyword parameter descriptions) and reword them for clarity and conciseness. I have tried to use docstring substitutions wherever applicable i.e. duplicated text.

(Leaving this open in draft form for now, as I fear I may forget about it otherwise, but continue to have more pressing work to address. There are ~20 other cases from a `git grep "If False then do not" | wc -l` yet to be tackled, to complete this PR, but since we are fosucing mostly on cf-python at the moment I doubt many merge conflicts will arise if I leave this for a while.)

### DIFF
--- a/cfdm/constructs.py
+++ b/cfdm/constructs.py
@@ -959,8 +959,8 @@ class Constructs(mixin.Container, core.Constructs):
         :Parameters:
 
             data: `bool`, optional
-                If False then do not copy data contained in the metadata
-                constructs. By default such data are copied.
+                If True (the default) then copy data contained in the
+                metadata constructs.
 
         :Returns:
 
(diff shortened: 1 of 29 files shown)


davidhassell (approved):
Thanks for this, Sadie - the docs are now much improved!

davidhassell:
Please do merge when you're ready.

sadielbartholomew:
One CI job has failed due to a spurious test failure (`FAIL: test_read_subsampled_coordinates (test_read_write.read_writeTest)`) which is unrelated and which I will look into separately. Approved and otherwise the tests (locally and the CI jobs) pass, so good to merge.

PR #184, opened 2022-05-04
catch missing Array

davidhassell:
`cfdm.Data.compressed_array` fails if there is no Array returned by `_get_Array`, which is OK for cfdm but not for daskified cf-python, for which there sill usually not be an Array object. This PR simply traps this case instead and raises the nice exception.


sadielbartholomew (approved):
A small, sensible and justified change that I can confirm doesn't break anything (and we can ignore the linting CI job failure which is failing due to a `black` internal issue - I might need to bump the version in the relevant workflow so I'll take a look at that separately).

PR #191, opened 2022-07-11
Read zero-length files

### LINKED ISSUE #190: `cfdm.read` hangs when reading a zero-length file
`cfdm.read` hangs when reading a zero-length file ...

```python
>>> import cfdm, subprocess
>>> subprocess.run(f"touch zero_length", shell=True, check=True)
>>> cfdm.read('zero_length')
# Hangs
```
... when it should raise an exception, such as `OSError: Can't determine format of file zero_length`.

```python
>>> import cfdm
Platform: Linux-5.4.0-121-generic-x86_64-with-glibc2.31
HDF5 library: 1.12.1
netcdf library: 4.8.1
Python: 3.9.5
netCDF4: 1.5.8
numpy: 1.22.3
cfdm.core: 1.9.0.3
cftime: 1.6.0
netcdf_flattener: 1.2.0
cfdm: 1.9.0.3
```

davidhassell:
Fixes #190

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,4 +1,13 @@
-Version 1.9.0.3
+Version 1.9.0.4
+---------------
+
+**2022-??-??**
+
+* Fix bug that caused a hang when reading zero-length files
+  (https://github.com/NCAS-CMS/cfdm/issues/190)
+
+----
+  Version 1.9.0.3
 ---------------
 
 **2022-03-10**
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -7,7 +7,8 @@ Version 1.9.0.4
   (https://github.com/NCAS-CMS/cfdm/issues/190)
 
 ----
-  Version 1.9.0.3
+
+Version 1.9.0.3
 ---------------
 
 **2022-03-10**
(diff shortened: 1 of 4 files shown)


Changelog.rst
```diff
@@ -1,4 +1,13 @@
...
+Version 1.9.0.4
+---------------
+
+**2022-??-??**
+
+* Fix bug that caused a hang when reading zero-length files
+  (https://github.com/NCAS-CMS/cfdm/issues/190)
+
+----
+  Version 1.9.0.3
```
sadielbartholomew, line 10: Looks like this got shifted accidentally (and we lost a space that might be required) which will probably break the formatting, so:
```suggestion

Version 1.9.0.3
```

sadielbartholomew (approved):
One minor comment, but otherwise a solid fix:

* I recreated the issue at hand on the current `master`, and observed that this branch resolves it;
* I checked this didn't cause adverse effects for the reading of other forms of file, notably given the fix applied, checking the logic for, and testing with CDL, that it didn't throw a wobbly for files with at least one blank line to begin with that could otherwise be read fine;
* the new test looks sensible and captures the original problem.

Thanks! Feel free to merge when ready.

PR #196, opened 2022-07-13
Under-the-hood changes for upstream dask implementation

davidhassell:


cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -18,6 +18,18 @@
 class NetCDFWrite(IOWrite):
     """A container for writing Fields to a netCDF dataset."""
 
+    def __new__(cls, *args, **kwargs):
+        """Store the NetCDFRead class.
+
+        .. note:: If a child class requires a different NetCDFRead class
+        than the one defined here, then it must be redefined in the
+        child class.
```
sadielbartholomew, line 26: Just thinking this comment is probably expressing something quite obvious? I guess you want to stress this, in which case it might be worth including, otherwise I'd be tempted to drop it?

cfdm/field.py
```diff
@@ -92,8 +92,8 @@ def __new__(cls, *args, **kwargs):
         """Store component classes.
 
         .. note:: If a child class requires a different component
-        classes           than the ones defined here, then they must be
-        redefined           in the child class.
+        classes than the ones defined here, then they must           be
```
sadielbartholomew, line 95: See my first comment with regards to whether such comments are worthwhile, and if so:

```suggestion
        classes than the ones defined here, then they must be
```

and there's an 'a' we should drop in the previous line for the 'classes' plural case.

sadielbartholomew (approved):
A few minor comments, but all good generally and I can see that this is resolving some trivially-fixed issues that the downstream cf-python `lama-to-dask` branch runs into across the full test suite, as per the PR motivation.

cfdm/field.py
davidhassell: Dropping this text, as per previous comment.

cfdm/read_write/netcdf/netcdfwrite.py
davidhassell: Fixed in https://github.com/NCAS-CMS/cfdm/pull/196/commits/8d8d6949d76351ac36090a86971a82eacb363ac2

sadielbartholomew (approved):
Feedback addressed, thanks. All good, merging so I can review the follow-on PR / for high throughput.

PR #203, opened 2022-07-27
`Field.indices`

### LINKED ISSUE #202: Create indices for Field constructs from metadata values
@JonathanGregory suggested that it would be natural in cfdm to be able to subspace field constructs based on metadata constructs values, rather than just by indices. An excellent idea.

For instance, with
```python
>>> import cfdm
>>> f = cfdm.example_field(0)
>>> print(f)
Field: specific_humidity (ncvar%q)
----------------------------------
Data            : specific_humidity(latitude(5), longitude(8)) 1
Cell methods    : area: mean
Dimension coords: latitude(5) = [-75.0, ..., 75.0] degrees_north
                : longitude(8) = [22.5, ..., 337.5] degrees_east
                : time(1) = [2019-01-01 00:00:00]
```

we currently can only create a new field for only longitude=337.5 by:
```python
>>> print(f[:, 7])
Field: specific_humidity (ncvar%q)
----------------------------------
Data            : specific_humidity(latitude(5), longitude(1)) 1
Cell methods    : area: mean
Dimension coords: latitude(5) = [-75.0, ..., 75.0] degrees_north
                : longitude(1) = [337.5] degrees_east
                : time(1) = [2019-01-01 00:00:00]
```
However, this requires knowledge of **a)** which position the longitude dimension is and **b)** where along the longitude axis `337.5` is.

It seems a natural use of CF metadata to be able to do this _without_ have to know a priori the answers to **a)** and **b)**.

cf-python does this with its `Field.indices`  and `Field.subspace` methods.

These methods are fully generalised, rely on the `cf.Query` class, and take cyclicity into account.

I don't propose that cfdm goes this far (certainly not yet!), so I suggest that we for cfdm we just introduce a `cfdm.Field.indices` method that behave much like its cf-python counterpart, but is restricted to
* Only 1-d metadata constructs (no domain axis constructs nor N-d constructs)
* Only metadata construct data values can be provided (as oposed to axis indices)
* No fancy "modes" (full, envelope, compress)  - cfdm doesn't have the machinery to back them up

The result would be:
```python
>>> ind = f.indices(longitude=337.5)
>>> print(ind)
(slice(None, None, None), array([False, False, False, False, False, False, False,  True]))
>>> print(f[ind])
Field: specific_humidity (ncvar%q)
----------------------------------
Data            : specific_humidity(latitude(5), longitude(1)) 1
Cell methods    : area: mean
Dimension coords: latitude(5) = [-75.0, ..., 75.0] degrees_north
                : longitude(1) = [337.5] degrees_east
                : time(1) = [2019-01-01 00:00:00]
```

The default position is to not burden cfdm with too much bloat (relative to cf-python), and cfdm started off very minimal. Since then there has been a trickle for back-ported functionality from cf-python to cfdm that has become obviously needed by people whilst actully using the library. There is no problem in doing this so long as each case is assessed on its own merits.

davidhassell:
Fixes #202

### DIFF
--- a/cfdm/mixin/fielddomain.py
+++ b/cfdm/mixin/fielddomain.py
@@ -1,6 +1,9 @@
 import logging
 import re
 
+import numpy as np
+
+# from ..data import Data
 from ..decorators import _manage_log_level_via_verbosity
 
 logger = logging.getLogger(__name__)
--- a/cfdm/mixin/fielddomain.py
+++ b/cfdm/mixin/fielddomain.py
@@ -1,9 +1,6 @@
 import logging
 import re
 
-import numpy as np
-
-# from ..data import Data
 from ..decorators import _manage_log_level_via_verbosity
 
 logger = logging.getLogger(__name__)
--- a/cfdm/test/test_Field.py
+++ b/cfdm/test/test_Field.py
@@ -6,7 +6,7 @@ import re
 import tempfile
 import unittest
 
-import numpy
+import numpy as np
 
 faulthandler.enable()  # to debug seg faults and timeouts
 
@@ -130,7 +130,7 @@ class FieldTest(unittest.TestCase):
                 indices = list(indices)
                 for axis, i in enumerate(indices):
                     if isinstance(i, list):
-                        e = numpy.take(e, indices=i, axis=axis)
+                        e = np.take(e, indices=i, axis=axis)
                         indices[axis] = slice(None)
 
                 e = e[tuple(indices)]
@@ -177,7 +177,7 @@ class FieldTest(unittest.TestCase):
     #            array = f[indices].data.array
     #            self.assertTrue((array == -1).all())
     #
-    #            values, counts = numpy.unique(f.data.array, return_counts=True)
+    #            values, counts = np.unique(f.data.array, return_counts=True)
     #            self.assertEqual(counts[0], array.size)
 
     def test_Field_get_filenames(self):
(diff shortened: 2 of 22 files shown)


davidhassell:
`cfdm.Field.indices` is pared-down code culled from `cf.mixin.fielddomain._indices`.

cfdm/mixin/fielddomain.py
```diff
@@ -1,6 +1,9 @@
 import logging
 import re
 
+import numpy as np
```
sadielbartholomew, line 4: Flake8 is saying:

```console
cfdm/mixin/fielddomain.py:4:1: F401 'numpy as np' imported but unused
```

(and I've double checked it is indeed so) so this can be removed (guessing it was left over from dev work).

sadielbartholomew (approved):
I agree with the approach suggested in #202 to address the issue at hand, and this implements it correctly, with sufficient new testing and documentation. I've raised a few minor comments, but overall this is good to merge, so do so at will. Thanks.

cfdm/mixin/fielddomain.py
davidhassell: Indeed it was. Fixed by https://github.com/NCAS-CMS/cfdm/pull/203/commits/8e5450c02d6b8093a6de1da6144f89c636a0d4cb

cfdm/test/test_Field.py
```diff
@@ -790,6 +790,41 @@ def test_Field_domain_axis(self):
...
+
+        with self.assertRaises(ValueError):
+            f.indices(longitude=-999)
+
+        with self.assertRaises(ValueError):
+            f.indices(longitude="bad_value_type")
+
+        key = f.construct("longitude", key=True)
+        with self.assertRaises(ValueError):
+            f.indices(longitude=112.5, key=22.5)
```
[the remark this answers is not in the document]
davidhassell: Fixed by https://github.com/NCAS-CMS/cfdm/pull/203/commits/8e5450c02d6b8093a6de1da6144f89c636a0d4cb

PR #205, opened 2022-07-27
Allow `array` and `datetime_array` attributes on data-bearing contructs

### LINKED ISSUE #204: Allow `array` and `datetime_array` attributes on data-bearing contructs
Currently, the only way to access an array of a construct that has data is `f.data.array`. Experience indicates that this is needlessly verbose, and `f.array` should also work. Similarly for `datetime_array`.

davidhassell:
Fixes #204

### DIFF
--- a/cfdm/test/test_DimensionCoordinate.py
+++ b/cfdm/test/test_DimensionCoordinate.py
@@ -3,7 +3,8 @@ import faulthandler
 import os
 import unittest
 
-import numpy
+import cftime
+import numpy as np
 
 faulthandler.enable()  # to debug seg faults and timeouts
 
@@ -19,7 +20,7 @@ class DimensionCoordinateTest(unittest.TestCase):
 
     dim = cfdm.DimensionCoordinate()
     dim.set_property("standard_name", "latitude")
-    a = numpy.array(
+    a = np.array(
         [
             -30,
             -23.5,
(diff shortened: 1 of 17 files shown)


cfdm/test/test_DimensionCoordinate.py
```diff
@@ -107,6 +108,21 @@ def test_DimensionCoordinate_climatology(self):
         self.assertTrue(x.del_climatology())
         self.assertIsNone(x.del_climatology(None))
 
+    def test_DimensiconCoordinate_array(self):
+        """Test the `DimensionCoordinate.datetime_array` method."""
```
sadielbartholomew, line 112:
```suggestion
        """Test the `DimensionCoordinate.array` method."""
```

cfdm/test/test_DimensionCoordinate.py
```diff
@@ -107,6 +108,21 @@ def test_DimensionCoordinate_climatology(self):
...
+        f = cfdm.example_field(0)
+        t = f.construct("time")
+        self.assertEqual(t.array, 31)
```
sadielbartholomew, line 115: For thoroughness, can we also include a check that these direct methods are effectively aliases for the equivalent `data.` method, i.e:

```suggestion
        self.assertEqual(t.array, 31)
        self.assertEqual(t.array, t.data.array)
```
?

sadielbartholomew (commented):
.

cfdm/test/test_DimensionCoordinate.py
```diff
@@ -107,6 +108,21 @@ def test_DimensionCoordinate_climatology(self):
...
+
+    def test_DimensiconCoordinate_datetime_array(self):
+        """Test the `DimensionCoordinate.datetime_array` method."""
+        f = cfdm.example_field(0)
+        t = f.construct("time")
+        self.assertEqual(
+            t.datetime_array,
+            np.array([cftime.DatetimeGregorian(2019, 1, 1, 0, 0, 0, 0)]),
+        )
```
sadielbartholomew, line 124: Equivalently to the previous comment, for ease of incorporating if you agree it would be good to add:

```suggestion
        )
        self.assertEqual(t.datetime_array, t.data.datetime_array)
```
```

PR #210, opened 2022-08-24
Disallow subspaces of constructs that have one or more size 0 axes

### LINKED ISSUE #209: Subspaces of constructs can currently have one or more size 0 axes
`numpy` (and similar data objects that share its API, such as `dask`, `cfdm.Data`) reasonably allow subspaces to have size 0 axes:
```python
>>> import numpy as np
>>> a = np.arange(9)
>>> a[[False] * 9]
array([], dtype=int64)
```
At cfdm `v1.10.0.0`, cfdm _constructs_ also allow this:

```python
>>> import cfdm
>>> f = cfdm.example_field(0)
>>> f.shape
(5, 8)
>>> f[[False] * 5]
<Field: specific_humidity(latitude(0), longitude(8)) 1>
```
... but they shouldn't, because it is not allowed by the CF data model. I.e. in the above example, `f[[False] * 5]` should raise an exception.

PR to follow.

davidhassell:
Fixes #209

### DIFF
--- a/cfdm/field.py
+++ b/cfdm/field.py
@@ -300,7 +300,7 @@ class Field(
         if 0 in new_data.shape:
             raise IndexError(
                 f"Indices {indices!r} result in a subspaced shape of "
-                f"{new_data.shape}, but can't create a subspace of a "
+                f"{new_data.shape}, but can't create a subspace of "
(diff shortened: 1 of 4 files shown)


sadielbartholomew (approved):
Perfect, thanks. Please merge.

PR #212, opened 2022-08-26
 Fix for failure when printing date-time data with the first element masked

### LINKED ISSUE #211: `Data.__str__` fails when the data contains date-times with the first element masked
At cfdm v1.10.0.0 we have:

```python
>>> # v1.10.0.0
>>> import cfdm
>>> import numpy as 
>>> dt = np.ma.array([10, 20], mask=[True, False])
>>> d = cfdm.Data(dt, units='days since 2000-01-01')
>>> str(d)
Traceback
    ...
TypeError: unsupported operand type(s) for +: 'cftime._cftime.DatetimeGregorian' and 'NoneType'
```
but what we expect is 
```python
>>> # Expected behaviour:
>>> d = cfdm.Data(dt, units='days since 2000-01-01')
>>> str(d)
'[--, 2000-01-21 00:00:00]'

```

davidhassell:
Fixes #211

### DIFF
--- a/cfdm/test/test_Data.py
+++ b/cfdm/test/test_Data.py
@@ -73,6 +73,7 @@ class DataTest(unittest.TestCase):
         dt = numpy.ma.array([10, 20], mask=[True, False])
         d = cfdm.Data(dt, units="days since 2000-01-01")
         self.assertTrue(str(d) == "[--, 2000-01-21 00:00:00]")
+        self.assertTrue(repr(d) == "<Data(2): [--, 2000-01-21 00:00:00]>")
 
     #    def test_Data__getitem__(self):
     def test_Data__setitem__(self):
--- a/cfdm/test/test_Data.py
+++ b/cfdm/test/test_Data.py
@@ -68,12 +68,6 @@ class DataTest(unittest.TestCase):
             _ = repr(d)
             _ = str(d)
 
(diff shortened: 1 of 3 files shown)


cfdm/test/test_Data.py
```diff
@@ -68,6 +68,12 @@ def test_Data__repr__str(self):
             _ = repr(d)
             _ = str(d)
 
+        # Test when the data contains date-times with the first
+        # element masked
+        dt = numpy.ma.array([10, 20], mask=[True, False])
+        d = cfdm.Data(dt, units="days since 2000-01-01")
+        self.assertTrue(str(d) == "[--, 2000-01-21 00:00:00]")
```
sadielbartholomew, line 75: Since the `repr` was also failing, might be good to also check that for robustness (I realise it probably calls `str()` in some way under-the-hood for the final component of the output, hence the corresponding failure, but for a one-liner perhaps why not)?

```suggestion
        self.assertTrue(str(d) == "[--, 2000-01-21 00:00:00]")
        self.assertTrue(repr(d) == "<Data(2): [--, 2000-01-21 00:00:00]>")
```

sadielbartholomew (approved):
Sensible fix to the issue at hand, with new testing to cover it. I've made one tiny suggestion, not remotely compulsory, so please merge when you are ready.

PR #216, opened 2022-09-09
Original filenames methods

### LINKED ISSUE #215: New "original filenames" methods
The conversation at https://github.com/NCAS-CMS/cf-python/issues/365 discussed how to record the names of files the contained some or all of the data at instantiation time - typically (but not `necessarily)` from a `read` call.

I propose a new suite of `original_filenames` methods that allow us to define, report, and edit these "original" files - those that contained some or all of the data at instantiation time - and are independent of the `get_filenames` methods. `get_filenames` reports on any files that are used by Data objects _at the time of asking_, which might reported different (typically fewer) files than were being used _at the time of data instantiation_.

It is also the case that the daskification of the cf-python library confuses functionality of the `get_filenames` methods - due to the new lazy execution paradigm, the notion of files in use _at the time of asking_ is a little fuzzy.

I propose along the lines of:

```python
>>> f = cfdm.read("parent_file.nc")[0]
>>> f.original_filenames()
{"parent_file.nc"}

>>> f = cfdm.read("parent_file.nc", external="external_file.nc")[0]
>>> f.original_filenames()
{"parent_file.nc", "external_file.nc"}
>>> f.original_filenames(update="file.nc")
>>> f.original_filenames()
{"parent_file.nc", "external_file.nc", "file.nc"}
>>> f.original_filenames(clear=True)
{"parent_file.nc", "external_file.nc", "file.nc"}
>>> f.original_filenames()
set()
```

You might want to edit the original file names if you are creating constructs ab initio (without `cfdm.read`, which will set them for you automatically), or if you are creating a Field (or Data or any other construct) that is some combination of other - in this case you might want to store the superset of original files.

Many thanks to @ThibHlln for first raising this over at the cf-python issue mentioned above. This functionally will be inherited by cf-python, so will be available there, too!

davidhassell:
Fixes #215


davidhassell:
OK - finished adding the bits I realised were missing!

cfdm/data/data.py
```diff
@@ -1901,6 +2046,150 @@ def minimum(self, axes=None):
...
+
+        .. versionadded:: (cfdm) 1.10.0.1
+
+        .. seealso:: `get_filenames`
+
+        :Parameters:
+
+            define: (sequence of) `str`, optional
+                Set these original file names, removing any already
+                stored.
```
sadielbartholomew, line 2075: Something to indicate explicitly that only one of the three keyword parameters can/should be set, i.e. not any two or all three, might be useful, even if it should be clear enough from the descriptions.

cfdm/test/test_read_write.py
```diff
@@ -886,6 +886,29 @@ def test_read_subsampled_coordinates(self):
...
+        x = f.dimension_coordinate("grid_longitude")
+
+        self.assertEqual(len(f.original_filenames()), 1)
+        self.assertEqual(len(x.original_filenames()), 1)
+
+        self.assertEqual(len(x.original_filenames(clear=True)), 1)
+        self.assertEqual(x.original_filenames(), set())
+
+        self.assertEqual(len(f.original_filenames(clear=True)), 1)
+        self.assertEqual(f.original_filenames(), set())
```
sadielbartholomew, line 901: Since we're just running the same assertions here for `x` and `f`, might be good to consolidate into a loop?

sadielbartholomew (approved):

sadielbartholomew (approved):
(To finalise the review started last week, with inline comments as raised above.)

Overall, I agree with the approach proposed in #215 to deal with the underlying deficiencies discussed in NCAS-CMS/cf-python#365, as applied here. It seems like a sensible way to deal with the issue at hand. **However, I did have some questions regarding the approach so please see https://github.com/NCAS-CMS/cf-python/issues/448#issuecomment-1263893466.**

As for the implementation, there are quite a few minor docs issues e.g. typos, as raised (and expected given the nature of the various new methods added and adaptations to the rest of the code required!), but otherwise it is sound (assuming the question referenced above is addressed), with an especially good job on the new tests which are nice and rigorous.

Please merge when ready. Thanks.

cfdm/test/test_read_write.py
davidhassell: Good idea. Done in https://github.com/NCAS-CMS/cfdm/pull/216/commits/d59240610d4538c3ab297dd58975b44b744a948f

cfdm/data/data.py
davidhassell: I'm happy leaving this to be trapped in the code:
```python
                raise ValueError(
                    "Can't set the 'define' and 'update' parameters "
                    "at the same time"
```
```python
                raise ValueError(
                    "Can't set the 'clear' parameter with either of the "
                    "'define' and 'update' parameters"
                )
```
Is that OK with you? Happy to doc it if you prefer.

sadielbartholomew: I'm inclined to say I think it should also be documented, even if that is to very concisely indicate as such, but it's your call!

davidhassell: OK - will do. I'll hold off implementing it until we're agreed on where we stand on https://github.com/NCAS-CMS/cf-python/issues/448#issuecomment-1266560596

davidhassell:
### Change of tack

Following a discussion with @sadielbartholomew on question raised at https://github.com/NCAS-CMS/cf-python/issues/448, the code has been refactored such that:

* defining, updating and clearing original filenames can only be done with the `_original_filenames` methods, which are available on all relevant classes
* original filename access is by the `get_original_filenames` methods, which are available on all relevant classes
* The on-demand link between a construct's original filenames and those of it's Data is broken.
* The original filenames are now stored on the constructs rather than their data (because some constructs don't have data!)
* It is, however, still convenient in `cfdm.read` to get original files from the data during the one-off initialisation. 

Ready for a new review!

Changelog.rst
```diff
@@ -3,6 +3,13 @@ Version 1.10.0.1
 
 **2022-??-??**
 
+* New method: `cfdm.Data.get_tie_point_indices`
+* New method: `cfdm.Data.get_interpolation_parameters`
+* New method: `cfdm.Data.get_dependent_tie_points`
+* Record the names of files that contain the original data
+  (https://github.com/NCAS-CMS/cfdm/issues/215)
+* New method: `cfdm.Field.original_filenames`
+* New method: `cfdm.Data.original_filenames`
```
sadielbartholomew, line 12: Needs updating to reflect the new approach:

```suggestion
* New method: `cfdm.Field.get_original_filenames`
* New method: `cfdm.Data.get_original_filenames`
```

sadielbartholomew (approved):
The changelog needs updating on account of the new approach in https://github.com/NCAS-CMS/cfdm/pull/216#issuecomment-1268566563 as raised inline, but overall the new commits adapt the PR to implement that precise approach, and do so in a very elegant manner (nice use of a new mixin class)! :100: Thanks and please merge when ready.

PR #218, opened 2022-10-09
Test get_original_filenames in `cfdm.write`

### LINKED ISSUE #192: Remove test that prevents writing to a file that is about to be read
There is currently a reasonable test in `cfdm.write` to prevent writing to any file that needs to be read to realise a Field construct in memory: https://github.com/NCAS-CMS/cfdm/blob/v1.9.0.3/cfdm/read_write/netcdf/netcdfwrite.py#L4305-L4312

```python
        if fields and mode != "r":
            filename = os.path.abspath(filename)
            for f in fields:
                if filename in self.implementation.get_filenames(f):
                    raise ValueError(
                        "Can't write to a file that contains data "
                        f"that needs to be read: {filename}"
                    )
```

This is fine so long as any other implementation can provide a `get_filenames()` method with the same API.

However, it is becoming apparent that the future daskified version of cf-python can't reliably provide a `get_filenames()` method.

Therefore, I propose simply removing the above `if` clause from cfdm. It'd be nice to not have to do this, but I can't see a way around it. 

Removing this test would mean that cfdm (and cf-python) would, at least, become consistent with netCDF4-python, in that a file opened for reading can get trashed if subsequently opened for writing.

What do you think?

davidhassell:
Fixes #192.

Following on from discussions around https://github.com/NCAS-CMS/cf-python/issues/454, I think it makes sense to change the test to one for `get_original_filenames`, rather than deprecating `get_filenames` at this time.

### DIFF
--- a/cfdm/cfdmimplementation.py
+++ b/cfdm/cfdmimplementation.py
@@ -1625,6 +1625,20 @@ class CFDMImplementation(Implementation):
         """
         return cell_measure.get_measure(default=None)
 
+    def get_original_filenames(self, parent):
+        """Get the original names of the files containing the construct.
+
+        :Parameters:
+
+            parent:
+
+        :Returns:
+
+            `set`
+
+        """
+        return parent.get_original_filenames()
+
     def nc_get_dimension(self, parent, default=None):
         """Return the netCDF variable name.
 
--- a/cfdm/cfdmimplementation.py
+++ b/cfdm/cfdmimplementation.py
@@ -1631,7 +1631,7 @@ class CFDMImplementation(Implementation):
         :Parameters:
 
             parent: Construct
-                The construct to query of roriginal filenames.
+                The construct to query of original filenames.
 
         :Returns:
 
--- a/cfdm/cfdmimplementation.py
+++ b/cfdm/cfdmimplementation.py
@@ -1630,11 +1630,13 @@ class CFDMImplementation(Implementation):
 
         :Parameters:
 
-            parent:
+            parent: Construct
+                The construct to query of roriginal filenames.
 
         :Returns:
 
             `set`
+                The original filenames
 
         """
         return parent.get_original_filenames()
--- a/cfdm/read_write/netcdf/netcdfwrite.py
+++ b/cfdm/read_write/netcdf/netcdfwrite.py
@@ -4289,7 +4289,11 @@ class NetCDFWrite(IOWrite):
                 `netCDF.Dataset` instance.
 
             fields: sequence of `Field` or `Domain`
-                The constructs to be written.
+                The constructs to be written to the netCDF file. Note
+                that these constructs are only used to ascertain if
+                any data to written be is in *filename*. If this is
+                the case and mode is "w" then an exception is raised
+                to prevent *filename* from being deleted.
 
         :Returns:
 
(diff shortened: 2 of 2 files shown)


cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -4288,8 +4288,8 @@ def file_open(self, filename, mode, fmt, fields):
                 As for the *format* parameter for initialising a
                 `netCDF.Dataset` instance.
 
-            fields: sequence of `Field`
-                The field constructs to be written.
+            fields: sequence of `Field` or `Domain`
+                The constructs to be written.
```
sadielbartholomew, line 4292: I realise you are just extending the description to include `Domain` here, but (in the spirit of raising issues as we see them) actually this parameter description overall seems pretty obscure to me, such that I had to check the code and try to deduce the point here (hoping it's not just a 'Monday Morning' type thing and it is in fact obvious!).

Mainly I was confused why a `file_open` method would be writing anything, as this might imply. So as I understand it from checking the code itself and usage of this, the constructs given in `field` are the ones the user *intends to* (eventually) write out to the file to be opened, though the method doesn't do any writing. I remember we planned to give this code a good refactor / rewrite when we had time, when we could change this to self-contained methods, but for now it would be good to amend the description to imply that sort of thing! Something like:

```suggestion
                The constructs that will at some point be
                written to the file once opened.
```

(Realising the above situation not ideal really as how would someone always know that in advance?)

Assuming I'm not completely mis-understanding this!

sadielbartholomew (approved):
> Following on from discussions around https://github.com/NCAS-CMS/cf-python/issues/454, I think it makes sense to change the test to one for get_original_filenames, rather than deprecating get_filenames at this time.

Agreed.

The PR is generally good, though I raised one minor comment and had a little ramble about something that isn't clear, please let me know what you think with respect to that and merge when ready.

cfdm/cfdmimplementation.py
```diff
@@ -1625,6 +1625,20 @@ def get_measure(self, cell_measure):
         """
         return cell_measure.get_measure(default=None)
 
+    def get_original_filenames(self, parent):
+        """Get the original names of the files containing the construct.
+
+        :Parameters:
+
+            parent:
```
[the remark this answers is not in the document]
davidhassell: See https://github.com/NCAS-CMS/cfdm/pull/218/commits/1b943314a956769130aff4e3d981f3e2bd56266b

cfdm/read_write/netcdf/netcdfwrite.py
davidhassell: You are spot on. See https://github.com/NCAS-CMS/cfdm/pull/218/commits/1aed9f4c46b63ceb88cf953c74be94ccc51e83fb

davidhassell:
Nothing Monday-morning-ish about your two comments. How about these last two commits?

cfdm/cfdmimplementation.py
```diff
@@ -1630,11 +1630,13 @@ def get_original_filenames(self, parent):
 
         :Parameters:
 
-            parent:
+            parent: Construct
+                The construct to query of roriginal filenames.
```
sadielbartholomew, line 1634:
```suggestion
                The construct to query of original filenames.
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -4288,22 +4288,26 @@ def file_open(self, filename, mode, fmt, fields):
                 As for the *format* parameter for initialising a
                 `netCDF.Dataset` instance.
 
-            fields: sequence of `Field`
-                The field constructs to be written.
+            fields: sequence of `Field` or `Domain`
+                The constructs to be written to the netCDF file. Note
+                that these constructs are only used to ascertain if
+                any data to written be is in *filename*. If this is
```
sadielbartholomew, line 4294:
```suggestion
                any data to be written is in *filename*. If this is
```

sadielbartholomew (approved):
Two minor comments, but previous feedback addressed well. Thanks. Good spot regarding post append mode clean up! Ready to merge otherwise.

PR #219, opened 2022-10-09
Refactor `Field.compress` to help child implementations

davidhassell:
This change, which has no effect on the functionality, is primarily so that the code plays nicely with cf-python+dask, but it's also probably faster, too!


sadielbartholomew (approved):
All good. Please merge!

> but it's also probably faster, too!

Seems likely indeed, though this reminds me that we should plan to do some profiling (according to an agreed strategy) for relevant code changes and modules in cf-python so we can be verify speed improvements rather than just speculating!

PR #222, opened 2022-10-28
Allow data to be omitted from netCDF files during `cfdm.write`

### LINKED ISSUE #221: Allow data to be omitted from netCDF files during `cfdm.write`
It would be useful to be able to not write to disk the data from selected construct types.

This would not affect the amount of netCDF variables and dimensions that are written to the file, nor the netCDF variables' attributes, but would not create data on disk for the requested variables thus resulting in a smaller file than it otherwise would have been.

When the new file is read then the data of these variables will be represented by an array of all missing data.

This will be useful for cases when there is a need to store a file's metadata structure for inspection, but the data values are immaterial.

davidhassell:
Fixes #221

### DIFF
--- a/cfdm/test/test_read_write.py
+++ b/cfdm/test/test_read_write.py
@@ -920,7 +920,7 @@ class read_writeTest(unittest.TestCase):
         self.assertFalse(g.construct("grid_latitude").array.count())
 
         # Check that a dump works
-        f.dump(display=False)
+        g.dump(display=False)
 
         cfdm.write(f, tmpfile, omit_data=("field", "dimension_coordinate"))
         g = cfdm.read(tmpfile)[0]
--- a/cfdm/test/test_read_write.py
+++ b/cfdm/test/test_read_write.py
@@ -905,6 +905,36 @@ class read_writeTest(unittest.TestCase):
             set((cfdm.abspath(parent_file), cfdm.abspath(external_file))),
         )
 
+    def test_write_omit_data(self):
+        """Test the `omit_data` parameter to `write`."""
+        f = cfdm.example_field(1)
+        cfdm.write(f, tmpfile)
+
+        cfdm.write(f, tmpfile, omit_data="all")
+        g = cfdm.read(tmpfile)
+        self.assertEqual(len(g), 1)
+        g = g[0]
+
+        # Check that the data are missing
+        self.assertFalse(g.array.count())
+        self.assertFalse(g.construct("grid_latitude").array.count())
+
+        cfdm.write(f, tmpfile, omit_data=("field", "dimension_coordinate"))
+        g = cfdm.read(tmpfile)[0]
+
+        # Check that only the field and dimension coordinate data are
+        # missing
+        self.assertFalse(g.array.count())
+        self.assertFalse(g.construct("grid_latitude").array.count())
+        self.assertTrue(g.construct("latitude").array.count())
+
+        cfdm.write(f, tmpfile, omit_data="field")
+        g = cfdm.read(tmpfile)[0]
+
+        # Check that only the field data are missing
+        self.assertFalse(g.array.count())
+        self.assertTrue(g.construct("grid_latitude").array.count())
+
 
 if __name__ == "__main__":
     print("Run date:", datetime.datetime.now())
--- a/cfdm/test/test_read_write.py
+++ b/cfdm/test/test_read_write.py
@@ -919,6 +919,9 @@ class read_writeTest(unittest.TestCase):
         self.assertFalse(g.array.count())
         self.assertFalse(g.construct("grid_latitude").array.count())
 
+        # Check that a dump works
+        f.dump(display=False)
+
         cfdm.write(f, tmpfile, omit_data=("field", "dimension_coordinate"))
         g = cfdm.read(tmpfile)[0]
 
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -10,6 +10,8 @@ Version 1.10.0.1
   (https://github.com/NCAS-CMS/cfdm/issues/215)
 * New method: `cfdm.Field.get_original_filenames`
 * New method: `cfdm.Data.get_original_filenames`
+* New keyword parameter to `cfdm.write`: ``omit_data``
+  (https://github.com/NCAS-CMS/cfdm/issues/221)
 * Fixed bug that caused a failure when printing date-time data with
   the first element masked
   (https://github.com/NCAS-CMS/cfdm/issues/211)
--- a/cfdm/cfdmimplementation.py
+++ b/cfdm/cfdmimplementation.py
@@ -760,6 +760,25 @@ class CFDMImplementation(Implementation):
         except KeyError:
             return None
 
+    def get_construct_type(self, variable):
+        """Return the construct type of a variable.
+
+        .. versionadded:: (cfdm) 1.10.0.1
+
+        :Parameters:
+
+            variable: object
+                The object to get a construct type from.
+
+        :Returns:
+
+            `str` or `None`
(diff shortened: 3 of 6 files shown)


davidhassell:
OK - I've finished fiddling - good to review, now, thanks.

cfdm/test/test_read_write.py
```diff
@@ -905,6 +905,36 @@ def test_read_original_filenames(self):
...
+        """Test the `omit_data` parameter to `write`."""
+        f = cfdm.example_field(1)
+        cfdm.write(f, tmpfile)
+
+        cfdm.write(f, tmpfile, omit_data="all")
+        g = cfdm.read(tmpfile)
+        self.assertEqual(len(g), 1)
+        g = g[0]
+
+        # Check that the data are missing
```
sadielbartholomew, line 918: Can we also do a quick `dump` on the re-read field `g` to further check that the missing data doesn't confuse a representation call? I say this mostly because I recall we have previously had a bug with representation when a field had missing data for constructs (see NCAS-CMS/cf-python#326) so having data not present on disk could cause similar issues as a representation edge case that would be good to test against. (I did check manually that `dump` works for all of the example fields when re-read in after being written out with `omit_data="all"`, so the tests should pass, but would be good to include for coverage!)

sadielbartholomew (commented):
This generally looks great, implementing the feature request at hand in an elegant way, and apart from one suggestion relating to testing (see in-line) I can't pick out anything to say in fault or suggestion, but I have noticed a few things in reviewing that I wanted to check were expected, in case they are issues that mean some tweaking is required.

Namely, to investigate the new keyword I did the following, essentially writing out all of the example fields with `omit_data="all"` set, and eyeballing the `dump` on each field when read back in to compare with the original.

Everything looked good, with the re-read fields showing masked data e.g. `[[--, ..., --]]` instead of `[[1.0, ..., 8.0]]` but the metadata reported otherwise being reported as identical, with some exceptions, as follows. All of these differences were noticed in the 6th and 7th fields:

* *example field 6*: the `Auxiliary coordinate: ncvar%z` construct gets re-labelled as `Auxiliary coordinate: altitude`, which corresponds to the standard name of the bounds, but I don't think there should be a change in the label like that? Explicitly I see:

    ```diff
    - Auxiliary coordinate: ncvar%z
    + Auxiliary coordinate: altitude
        Geometry: polygon
        Bounds:axis = 'Z'
        Bounds:standard_name = 'altitude'
        Bounds:units = 'm'
        Bounds:Data(cf_role=timeseries_id(2), 3, 4) = [[[1.0, ..., --]]] m
        Interior Ring:Data(cf_role=timeseries_id(2), 3) = [[0, ..., --]]
    ```

    where I am also wondering why the first few Data points remain for the bounds and interior ring, i.e. why they don't change and show as `[[--, ..., --]]` like the other data that gets omitted?
* *example field 7*: the `_FillValue = -1073741824.0` changes slightly to `_FillValue = -1073741800.0`, is that expected and/or harmless?
* *generally* is it expected that datetimes are reported (i.e. does this indicate an under-the-hood issue, not just a representation quirk) after `omit_data` and re-reading as e.g. `Data(time(1)) = [0]`? And for multiple datetimes, they sometimes have a `0` entry e.g. (from example field 7 construct `Dimension coordinate: time`), `Data(time(3)) = [--, 0, --] gregorian`, though sometimes not e.g. `Bounds:Data(time(3), 2) = [[--, ..., --]] gregorian` from the same construct?

davidhassell:
> Data(time(1)) = [0]? 

This one's an unrelated bug. If we look at the actual array we get:
```python
>>> t = g.construct('time')
>>> t.data
<Data(3): [--, 0, --] gregorian>   # Wrong
>>> t.data.array 
masked_array(data=[--, --, --],    # Right
             mask=[ True,  True,  True],
       fill_value=1e+20,
            dtype=float64)
```
So the underlying data is correct, but the `str` (which is called by the `repr` representation is wrong. Let's raise another issue for that.

davidhassell:
> example field 7: the _FillValue = -1073741824.0 changes slightly to _FillValue = -1073741800.0, is that expected and/or harmless?

Unexpected, but harmless for now! It happens at through a write/read cycle at v1.10.0.0, too, so isn't related to `omit_data`. Might be a rounding concern (I also looked at some double precision cases). One for another issue, too.

davidhassell:
> example field 6: the Auxiliary coordinate: ncvar%z construct gets re-labelled as Auxiliary coordinate: altitude, 

Again, unrelated (thankfully). If you inspect `f` and `g` after `import cfdm; f = cfdm.example_field(6); cfdm.write(f, 'delme1.nc'); cfdm.write(f, 'delme2.nc', omit_data='all'); f = cfdm.read('delme1.nc')[0]; g = cfdm.read('delme2.nc')[0]`, where `f` is read from disk, rather than created ab initio, all is fine. 

I'll have a look in the example field definition to see if it's obvious why "altitude" doesn't appear in the first place ...

sadielbartholomew:
Thanks David for drilling down to investigate all of the facets. Always good to catch a bug even if it isn't immediately related :smile:

So as far as I can tell from your recent comments, https://github.com/NCAS-CMS/cfdm/pull/222#issuecomment-1295175407 is the only issue directly related to this PR, and you have put in a fix, is that right? In which case I just need to re-review with the new commit and then we're good to go?

davidhassell:
> example field 6: the Auxiliary coordinate: ncvar%z construct gets re-labelled as Auxiliary coordinate: altitude,

This turns out to be a feature! It's to do with the unusual case where a coordinate construct has bounds but no data. In this case we shouldn't be assigning an `ncvar` to the coordiantes (which is what example 6 was doing). This commit https://github.com/NCAS-CMS/cfdm/pull/222/commits/3ae74fe0916df87256843ae1ea2017f9b2b54fd5 makes things consistent.

davidhassell:
> Can we also do a quick dump on the re-read field g

I've added to the test:
```python
        # Check that a dump works
        g.dump(display=False)
```
Is that what you meant?

sadielbartholomew:
> Is that what you meant?

Yep, precisely, just test that it doesn't fail miserably due to some facet of missing data.

davidhassell:
Great - are we good to go on this? (notwithstanding the two new issues that have been spawned :))

davidhassell:
(sorry - I missed https://github.com/NCAS-CMS/cfdm/pull/222#issuecomment-1295236660, but I think we've ended up in the same place!)

sadielbartholomew:
> (sorry - I missed https://github.com/NCAS-CMS/cfdm/pull/222#issuecomment-1295236660, but I think we've ended up in the same place!)

No worries! I also missed your comments afterwards so my apologies too.

I'm just taking a final look and will let you know in a moment.

cfdm/test/test_read_write.py
sadielbartholomew: Added in afebbd6 and the previous commit, so marking as resolved.

sadielbartholomew (approved):
All questions I raised in first review have been answered and problems emerging from those have either been addressed by new commits that I've since reviewed, or delegated to (intended) new issues (or identified as a feature! :laughing:):

> ... notwithstanding the two new issues that have been spawned :) ...

so we just need to open these two (https://github.com/NCAS-CMS/cfdm/pull/222#issuecomment-1295197082 and https://github.com/NCAS-CMS/cfdm/pull/222#issuecomment-1295208100), possibly just by linking to your comments, since I can't see any issues referenced or on the Issue Trackers so assume they haven;t been created yet. Shall I do that or would you like to?

All good, though! Thanks. Please merge.

davidhassell:
Thanks, Sadie - Happy for you to raise new issues :) (but just ping me otherwise)

PR #232, opened 2022-11-29
`get_filename` methods on all `Array` classes.

davidhassell:
Passes `get_filename` down to all `Array` objects, mainly so that it accessible to cf-python's need for deep navel-gazing.

### DIFF
--- a/cfdm/data/abstract/compressedarray.py
+++ b/cfdm/data/abstract/compressedarray.py
@@ -400,6 +400,28 @@ class CompressedArray(Array):
         """
         return {"data": self.source().copy()}
 
+    def get_filenames(self):
+        """Return the names of any files containing the compressed data.
+
+        .. versionadded:: (cfdm) 1.10.0.2
+
+        :Returns:
+
+            `set`
+                The file names in normalised, absolute form. If the
+                data are all in memory then an empty `set` is
+                returned.
+
+        """
+        data = self._get_compressed_Array(None)
+        if data is None:
+            return set()
+
+        try:
+            return data.get_filenames()
+        except AttributeError:
+            return set()
+
     def get_Subarray(self):
         """Return the Subarray class.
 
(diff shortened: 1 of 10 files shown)


cfdm/data/abstract/compressedarray.py
```diff
@@ -400,6 +400,28 @@ def conformed_data(self):
...
+
+        """
+        data = self._get_compressed_Array(None)
+        if data is None:
+            return set()
+
+        try:
+            return data.get_filenames()
+        except AttributeError:
+            return set()
```
sadielbartholomew, line 423: It looks like only `get_filename` can produce an `AttributeError` now, with the plural equivalents all catching those and returning the empty set (and I double checked this includes `data.get_filenames`), so:

```suggestion
        return data.get_filenames()
```

is sufficient, unless I am missing something?

cfdm/data/mixin/arraymixin.py
```diff
@@ -137,6 +139,63 @@ def get_compression_type(self):
         """
         return self._get_component("compression_type", "")
 
+    def get_filename(self, default=AttributeError()):
+        """The name of the file containing the array.
+
+        If there are multiple files then an `AttributeError` is
+        raised.
```
sadielbartholomew, line 146:
```suggestion
        raised by default.
```

Just to avoid any confusion RE the existence of the `default` parameter.

sadielbartholomew (approved):
I've raised a few minor comments, but otherwise this is all good and ready to merge, so please do so when ready. Thanks.

cfdm/data/abstract/compressedarray.py
davidhassell: Hi Sadie, The `AttributeError` here is to catch the case that the underlying compressed array does not support the `get_filenames` API (who knows what something like cf-python might put there :)), so I think that we still need it ...

sadielbartholomew: Oh fair, that makes sense. Thanks for clarifying, marking as resolved.

PR #237, opened 2023-01-06
Improve source docstrings in `__init__` methods

### LINKED ISSUE #236: Improve source docstrings in `__init__` methods
The `source` parameter of all `__init__` methods says something along the lines of "Initialize the parameters from those of *source*" which, at best, is a little vague. It would be good to provide a bit more detail on what this really means.

See https://github.com/NCAS-CMS/cf-python/issues/492 for more context and more detail.

I propose adding the same boilerplate text to each class. E.g. for `cfdm.Field`:

```
            source: optional
                Initialise the `Field` instance from the *source*
                object. All other parameters, apart from *copy*, are
                ignored and their values are instead inferred from
                *source* using the `Field` API. Any parameters
                that can not be retrieved from *source* are assumed to
                have their default value.

                Note that if *source* is a `Field` instance then
                ``cfdm.Field(source=source)`` is equivalent
                to ``source.copy()``.
```

davidhassell:
Fixes #236 

The real substance is found in `cfdm/core/docstring/docstring.py`, and nearly all of the other files changes are just rolling this out.

### DIFF
--- a/cfdm/core/docstring/docstring.py
+++ b/cfdm/core/docstring/docstring.py
@@ -64,15 +64,13 @@ _docstring_substitution_definitions = {
     # init properties
     "{{init properties: `dict`, optional}}": """properties: `dict`, optional
                 Set descriptive properties. The dictionary keys are
-                property names, with corresponding values. Ignored if
-                the *source* parameter is set.
+                property names, with corresponding values.
 
                 Properties may also be set after initialisation with
                 the `set_properties` and `set_property` methods.""",
     # init data
     "{{init data: data_like, optional}}": """data: data_like, optional
-                Set the data. Ignored if the *source* parameter is
-                set.
+                Set the data.
 
                 {{data_like}}
 
@@ -80,15 +78,13 @@ _docstring_substitution_definitions = {
                 `set_data` method.""",
     # init bounds
     "{{init bounds: `Bounds`, optional}}": """bounds: `Bounds`, optional
-                Set the bounds array. Ignored if the *source*
-                parameter is set.
+                Set the bounds array.
 
                 The bounds array may also be set after initialisation
                 with the `set_bounds` method.""",
     # init geometry
     "{{init geometry: `str`, optional}}": """geometry: `str`, optional
-                Set the geometry type. Ignored if the *source*
-                parameter is set.
+                Set the geometry type.
 
                 The geometry type may also be set after initialisation
                 with the `set_geometry` method.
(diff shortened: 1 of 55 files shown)


docs/Makefile
```diff
@@ -11,7 +11,7 @@ PAPER         =
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
 #ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) source
-ALLSPHINXOPTS   = -j 2 source
+ALLSPHINXOPTS   = -j 1 source
```
sadielbartholomew, line 14: Was this an intentional change? If so, what is the purpose of it?

sadielbartholomew (commented):
All good, generally, though I have raised a few comments/questions.

docs/Makefile
davidhassell: Pretty arbitrary: By default we run the sphinx build with 2 PEs. but sometimes (on my laptop) that's slower than with just one.

cfdm/core/docstring/docstring.py
```diff
@@ -97,18 +93,25 @@
...
-    "{{init source}}": """Note that if *source* is a `{{class}}` instance then
+    "{{init source: optional}}": """source: optional
+                Initialise the `{{class}}` instance from the *source*
+                object. All other parameters, apart from *copy*, are
+                ignored and their values are instead inferred from
+                *source* using the `{{class}}` API. Any parameters
+                that can not be retrieved from *source* are assumed to
+                have their default value.
+
+                Note that if *source* is a `{{class}}` instance then
```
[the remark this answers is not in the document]
davidhassell: How about https://github.com/NCAS-CMS/cfdm/pull/237/commits/f23bcdd438299e619add6e623d47860f553d2715?

docs/Makefile
sadielbartholomew: Fair enough, that's fine by me! Thought I'd check since the change seemed unrelated to the core of the PR...

cfdm/core/docstring/docstring.py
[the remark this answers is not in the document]
sadielbartholomew: That's brilliant, thanks - does what I hoped for RE the above in a very concise way! Marking as resolved.

sadielbartholomew (approved):
All questions answered and the one piece of feedback has been addressed well. One typo remains but after nuking that please merge!

PR #241, opened 2023-02-14
`del_properties` methods, and `ignore_properties` keyword to `cf.unique_constructs`

### LINKED ISSUE #240: New `ignore_properties` keyword to `cfdm.unique_constructs`
It would be useful to allow  `cfdm.unique_constructs` to ignore selected properties when assessing equality of input constructs. For example, two `Field` constructs may differ only by a timestamp within their "history" properties, and this may be of no concern. This would look like:

```python
>>> import cfdm
>>> f = cfdm.example_field(0)
>>> f2 = f.copy()
>>> f.set_property('foo', 'bar')
>>> f2.set_property('foo', 'bar2')
>>> cfdm.unique_constructs([f, f2]) 
[<Field: specific_humidity(latitude(5), longitude(8)) 1>,
 <Field: specific_humidity(latitude(5), longitude(8)) 1>]
>>> cfdm.unique_constructs([f, f2], ignore_properties='foo')
[<Field: specific_humidity(latitude(5), longitude(8)) 1>]

### PR DESCRIPTION
**Pull request by davidhassell**
Fixes #240

### CODE REVIEW

**Comment by sadielbartholomew** [comment]
Hi David, since you haven't merged are you wanting and awaiting a review on this, and then for NCAS-CMS/cf-python#598 too? If so I will assign myself and get it done shortly.

**Comment by davidhassell** [comment]
Hi Sadie - thanks for asking, but I think it's best to hold off for now, as it turns out Grenville is still testing it. I'll request the review formally when he's done (in a week or two?).

**Comment by sadielbartholomew** [comment]
Oh sure, I wasn't aware of that so thanks for the clarification. I'll wait until it pops up in my notifications.

**Comment by davidhassell** [comment]
Ready now - sooner than expected. See also https://github.com/NCAS-CMS/cf-python/pull/598.

File: cfdm/mixin/fielddomain.py [hunk 1]
```diff
@@ -1853,7 +1853,12 @@ def equals(
 
         """
         # Check the properties and data
-        ignore_properties = tuple(ignore_properties) + ("Conventions",)
+        if not ignore_properties:
+            ignore_properties = ("Conventions",)
```
sadielbartholomew, line 1857: Since the "Conventions" property is always ignored, as clearly stated in the docstring parameter description, is it slightly misleading to have a default of `None` for the keyword when we always ignore that anyway? Could/should we change the default to `("Conventions",)`? (Just throwing that question out there, not sure where I sit myself.)

sadielbartholomew (approved):
Some minor comments raised, but otherwise all good, with the new testing nice and thorough. Please merge when ready.

cfdm/functions.py
```diff
@@ -443,14 +443,29 @@ def unique_constructs(constructs, copy=True):
...
+
+            If *ignore_properties* has been set then *copy* is ignored
+            and deep copies are always returned, even if
+            *ignore_properties* is an empty sequence.
 
     :Returns:
 
-        `list`
-            The unique constructs. May be an empty list.
+        Sequence of constructs
```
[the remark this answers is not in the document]
davidhassell: Hi Sadie - I think that "sequence" includes generator, here.

cfdm/mixin/fielddomain.py
davidhassell: Hi - not sure. If the default was `("Conventions",)` then it would imply that you can remove it, which you can't, and you would have to add it in manually if you set your own property.

cfdm/functions.py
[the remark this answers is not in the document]
sadielbartholomew: Fair enough, I guess! Marking as resolved.

cfdm/mixin/fielddomain.py
sadielbartholomew: Oh, so if you can't remove it (context I didn't know though I probably should know, sorry :grimacing:) then I agree we should leave the default and set up as it is. Marking as resolved!

sadielbartholomew (approved):
All minor feedback addressed, so good to go and merge (in case a re-confirmation is helpful).

PR #247, opened 2023-03-03
Allow file missing data indicator values to be stored on `NetCDFArray` instances

### LINKED ISSUE #246: Allow file missing data indicator values to be stored on `NetCDFArray` instances
It can be useful to applications to know the missing data indicator without opening the netCDF file. This is the case for active storage reductions currently being implemented in cf-python.

It is almost always the case that the netCDF file is open and available when the `NetCDFArray` instance is created, so it is easy for these to be stored on the object if desired, e.g. 

```python
>>> n = NetCDFArray(
...     'file.nc', 'ncvar', missing_values={"missing_value": -999.0, "valid_range": (-111, 222)}
... )
```

davidhassell:
Fixes #246

### DIFF
--- a/cfdm/data/netcdfarray.py
+++ b/cfdm/data/netcdfarray.py
@@ -25,6 +25,7 @@ class NetCDFArray(abstract.Array):
         mask=True,
         units=False,
         calendar=False,
+        missing_values=None,
         source=None,
         copy=True,
     ):
@@ -107,6 +108,13 @@ class NetCDFArray(abstract.Array):
 
                 .. versionadded:: (cfdm) 1.10.0.1
 
+            missing_values: `dict`, optional
+                The missing value indicators defined by the netCDF
+                variable attributes. See `get_missing_values` for
+                details.
+
+                .. versionadded:: (cfdm) 1.10.0.3
+
             {{init source: optional}}
 
                 .. versionadded:: (cfdm) 1.10.0.0
@@ -164,6 +172,11 @@ class NetCDFArray(abstract.Array):
             except AttributeError:
                 calendar = False
 
+            try:
+                missing_values = source._get_component("missing_values", None)
+            except AttributeError:
+                missing_values = None
+
         if shape is not None:
             self._set_component("shape", shape, copy=False)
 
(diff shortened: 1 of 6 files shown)


cfdm/data/netcdfarray.py
```diff
@@ -454,6 +472,41 @@ def get_mask(self):
...
+        None
+
+        >>> a.get_missing_values()
+        {}
+
+        >>> a.get_missing_values()
+        {'missing_value': 1e20, 'valid_range': (-10, 20)}
+
+        >>> a.get_missing_values()
+        {'valid_min': -999}
```
sadielbartholomew, line 501: A bit pedantic, but if we're not indicating any change in `a` it might avoid any potential confusion if we use different variable names for each example, e.g:

```suggestion
        >>> a.get_missing_values()
        None

        >>> b.get_missing_values()
        {}

        >>> c.get_missing_values()
        {'missing_value': 1e20, 'valid_range': (-10, 20)}

        >>> d.get_missing_values()
        {'valid_min': -999}
```

cfdm/test/test_NetCDFArray.py
```diff
@@ -0,0 +1,79 @@
...
+        """Test NetCDFArray.get_missing_values."""
+        f = cfdm.example_field(0)
+
+        f.set_property("missing_value", -999)
+        f.set_property("_FillValue", -3)
+        f.set_property("valid_range", [-111, 222])
+        cfdm.write(f, tmpfile)
+
+        g = cfdm.read(tmpfile)[0]
+        print(g.data.source().get_missing_values())
```
sadielbartholomew, line 58:
```suggestion
        g.data.source().get_missing_values()
```

so the test doesn't produce output which clogs the STDOUT report, unless you deliberately want to check the `__str__` representation?

sadielbartholomew (approved):
Implements the new feature well, with good relevant new test coverage. I've raised some very minor comments; once you are happy, please merge (we can ignore the linting job failure, it is a flake8 update related workflow setup issue I have fixed on cf-python but need to sort out here and on cfunits).

cfdm/test/test_NetCDFArray.py
davidhassell: Ah - I think delete altogether. Good spot. https://github.com/NCAS-CMS/cfdm/pull/247/commits/41ad49c45273706c7d3e81d4c9ebbfc3a2bfbebc

PR #248, opened 2023-03-08
Minor updates to config. files (fixes pre-commit workflow)

sadielbartholomew:
Equivalent to NCAS-CMS/cf-python#588 so see the opening comment description there. Should fix the lintig workflow setup errors and genuine failures, which have been there for a little while on recently pushed commits.

Again, trivial setup changes therefore no peer review required.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,14 +3,15 @@
 # this configuration, simply execute 'flake8' under the root of the repo.
 
 [flake8]
-# Ignores should include listing of pycodestyle ignores as in test_style.py
+# Ignores should include listing of pycodestyle ignores as in test_style.py.
+# Note W503 is incompatible with PEP8 which we want to abide by
 ignore =
     W605,
     E272,
     E402,
     E501,
     E203,
-    W503  # Incompatible with PEP8 which we want to abide by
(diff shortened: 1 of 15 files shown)


sadielbartholomew:
@davidhassell FYI there seems to be one recurrent test method failure (looking at the recent commit history too), unrelated to any set up as far as I can tell:

```
======================================================================
FAIL: test_read_subsampled_coordinates (test_read_write.read_writeTest)
Test the reading subsampled coordinates.
CFDM TEST SUITE
---------------
Run date: 2023-03-08 15:18:25.445795
Platform: macOS-10.15.7-x86_64-i386-64bit 
  File "/Users/runner/work/cfdm/cfdm/cfdm/test/test_read_write.py", line 885, in test_read_subsampled_coordinates
    self.assertTrue(
AssertionError: False is not true

Ran 258 tests in 148.871s

FAILED (failures=1, skipped=1)
```

I'm not sure if you know of any recent changes that might have led to this? We should look into this separately.

For now, I can merge, as the workflows are updated as desired and the linting workflow is fixed as evidenced by the passing job.

PR #250, opened 2023-03-15
Fix error from `Data.second_element` for some data shapes 

### LINKED ISSUE #249: Error from `Data.second_element` for some data shapes
For some `Data` shapes, the retrieval of the second element causes an error:

```python
>>> import cfdm
>>> import numpy as np
>>> d = cfdm.Data(np.arange(6).reshape(1, 3, 2))  # This shape is OK
>>> d.array
array([[[0],
        [1],
        [2],
        [3],
        [4],
        [5]]])
>>> d.second_element()
1
>>> d = cfdm.Data(np.arange(6).reshape(1, 6, 1))  # Not OK
>>> d.second_element()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[51], line 1
----> 1 d.second_element()

File ~/cfdm/cfdm/data/data.py:3009, in Data.second_element(self)
   2981 def second_element(self):
   2982     """Return the second element of the data as a scalar.
   2983 
   2984     .. versionadded:: (cfdm) 1.7.0
   (...)
   3007 
   3008     """
-> 3009     return self._item(
   3010         (slice(0, 1, 1),) * (self.ndim - 1) + (slice(1, 2, 1),)
   3011     )

File ~/cfdm/cfdm/data/data.py:655, in Data._item(self, index)
    652 array = self[index].array
    654 if not numpy.ma.isMA(array):
--> 655     return array.item()
    657 mask = array.mask
    658 if mask is numpy.ma.nomask or not mask.item():

ValueError: can only convert an array of size 1 to a Python scalar
```

```python
Platform: Linux-5.15.0-67-generic-x86_64-with-glibc2.35
HDF5 library: 1.12.1
netcdf library: 4.8.1
Python: 3.10.9
netCDF4: 1.6.0
numpy: 1.22.3
cfdm.core: 1.10.0.3
cftime: 1.6.2
netcdf_flattener: 1.2.0
cfdm: 1.10.0.3
```

davidhassell:
Fixes #249

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,13 @@
+Version 1.10.?.?
+----------------
+
+**2023-??-??**
+
+* Fixed bug that caused `cf.Data.second_element` to fail for some data
+  shapes (https://github.com/NCAS-CMS/cfdm/issues/249)
+
+----
+
 Version 1.10.0.3
 ----------------
 
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -4,7 +4,8 @@ Version 1.10.?.?
 **2023-??-??**
 
 * Fixed bug that caused `cf.Data.second_element` to fail for some data
-  shapes (https://github.com/NCAS-CMS/cfdm/issues/249)
+  shapes, namely for a final axis with size one.
+  (https://github.com/NCAS-CMS/cfdm/issues/249)
 
 ----
 
--- a/cfdm/data/data.py
+++ b/cfdm/data/data.py
@@ -3006,9 +3006,7 @@ class Data(Container, NetCDFHDF5, Files, core.Data):
         bar <class 'str'>
 
         """
-        return self._item(
-            (slice(0, 1, 1),) * (self.ndim - 1) + (slice(1, 2, 1),)
-        )
+        return self._item(np.unravel_index(1, self.shape))
 
     @_inplace_enabled(default=False)
     def to_memory(self, inplace=False):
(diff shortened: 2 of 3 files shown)


Changelog.rst
```diff
@@ -1,3 +1,13 @@
+Version 1.10.?.?
+----------------
+
+**2023-??-??**
+
+* Fixed bug that caused `cf.Data.second_element` to fail for some data
+  shapes (https://github.com/NCAS-CMS/cfdm/issues/249)
```
sadielbartholomew, line 7: In my understanding of the original issue and fix, specifically the issue occurs when the final axes has a singular size i.e. last value in the shape is `1` (for quick testing on this, I used this):

```python
def test_shape(size, shape):
    d= cfdm.Data(np.arange(size).reshape(*shape))
    return d.second_element()

# And checking, e.g:
test_shape(60, (1, 15, 2, 2))  # OK
test_shape(60, (2, 15, 2, 1))  # Errors
# Etc.
```

so perhaps we can be more specific in case users think they have to re-check work or something drastic (in the unlikely case they even read the changelog!). I'm not sure if this is the clearest wording, but something like:

```suggestion
  shapes, namely for a final axis with size one.
  (https://github.com/NCAS-CMS/cfdm/issues/249)
```

sadielbartholomew (approved):
Fixes the issue at hand in a sensible way and adds some testing. I've made one minor suggestion, but all is good, so please merge when ready.

PR #252, opened 2023-03-24
Make `cfdm.read` always return unicode strings from netCDF files

### LINKED ISSUE #251: Make `cfdm.read` always return unicode strings from netCDF files
Currently, `cfdm.read` can return either unicode or byte strings, which sometimes makes testing for equality difficult, as the `equals` methods treat them the two data types as unequal (as does numpy).

It would be good it it was consistent, and the obvious choice is unicode.

```python
Platform: Linux-5.15.0-67-generic-x86_64-with-glibc2.35
HDF5 library: 1.12.1
netcdf library: 4.8.1
Python: 3.10.9
netCDF4: 1.6.0
numpy: 1.22.3
cfdm.core: 1.10.0.3
cftime: 1.6.2
netcdf_flattener: 1.2.0
cfdm: 1.10.0.3
```

davidhassell:
Fixes #251

### DIFF
--- a/cfdm/data/netcdfarray.py
+++ b/cfdm/data/netcdfarray.py
@@ -265,13 +265,14 @@ class NetCDFArray(abstract.Array):
             # A netCDF string type scalar variable comes out as Python
             # str object, so convert it to a numpy array.
             # --------------------------------------------------------
-            array = numpy.array(array, dtype=f"S{len(array)}")
+            array = numpy.array(array, dtype=f"U{len(array)}")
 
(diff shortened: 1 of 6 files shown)


sadielbartholomew:
@davidhassell are you looking for a review on this?

sadielbartholomew (approved):
> It would be good it it was consistent, and the obvious choice is unicode.

I agree with this approach which seems very sensible.

Implementation seems perfect. Please merge!

PR #257, opened 2023-05-31
Scrollable Navigation Menu (Sidebar)

### LINKED ISSUE #50: Docs: immobile navigation menu could be cut off
The navigation menu (left pane) of the documentation seems to have gotten a little longer with some new sections, & with the screen size of my (work) laptop has become just long enough to extend off the screen vertically:

cfdm-menu-long-scr

Nothing important is cut off at the moment with my screen view as an example, but whilst I know that is close to the end of the menu, users may not & may be frustrated at not being able to access the later parts of the menu without difficulty (zoom out etc.). It is not possible to scroll as there is no scrollbar (which is a style choice I agree with) & even forcing it by highlighting the text as works with the readthedocs theme doesn't work in this theme.

To alleviate potential for such problems, I suggest move some of the top-level headings under a more general page? 'Subclassing', 'Philosophy', 'Performance' & 'Versioning' are good candidates as they are all small pages & aren't especially standard docs sections. Maybe the former three could live under an umbrella 'Implementation' heading to go after the 'CF Data Model' section in listing?

Overall there are various approaches that would solve this, but I personally would rather not add a scrollbar to the nav menu, or change to a nested tree type of menu, as the menu looks very clean & clear as it is (just a little too long).

bewithankit:
Might resolve #50.

### DIFF
--- a/docs/_static/alabaster.css
+++ b/docs/_static/alabaster.css
@@ -30,8 +30,6 @@ div.sphinxsidebar {
     width: 220px;
     font-size: 14px;
     line-height: 1.5;
-    max-height: 100%;
-    overflow-y: auto;
 }
 
 hr {
--- a/docs/_static/alabaster.css
+++ b/docs/_static/alabaster.css
@@ -30,6 +30,8 @@ div.sphinxsidebar {
     width: 220px;
     font-size: 14px;
     line-height: 1.5;
+    max-height: 100%;
+    overflow-y: auto;
 }
 
 hr {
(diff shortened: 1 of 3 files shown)


docs/_static/alabaster.css
```diff
@@ -30,6 +30,8 @@ div.sphinxsidebar {
     width: 220px;
     font-size: 14px;
     line-height: 1.5;
+    max-height: 100%;
+    overflow-y: auto;
```
sadielbartholomew, line 34: As done in the equivalent PR to cf-python, it is better to put these in the `docs/source/_static/customise-alabaster.css` rather than the `alabaster.css` theme CSS, which to prevent the muddling of original theme and the customisation of it, should remain aligned with the basic Alabaster theme styling.

bewithankit: Thanks Sadie, I didn't spot that!

sadielbartholomew (approved):
Thanks Ankit. Fixes the issue at hand, so please merge.

PR #261, opened 2023-06-19
Fix very slow reads of datasets with compession by gathering or DSG ragged arrays

### LINKED ISSUE #260: Very slow reads of datasets with compession by gathering or DSG ragged arrays
In versions `1.10.0.1` to `1.10.1.0`, reading datasets with compression by gathering or DSG ragged arrays is very slow, taking seconds rather than milliseconds.

This is due to a bug in `cfdm.Data._original_filenames`, which causes every integer element in the list/count/index variable to be assessed for an original filename, which, of course, it does not have!

PR to follow.

davidhassell:
Fixes #260 

Also includes the `compression_index` keyword parameter to `NetCDFRead._create_data`. This is not used by cfdm, yet, but will be by cf-python.


sadielbartholomew (approved):
Fixes the issue at hand in a logical way. Relevant tests in the test suite (and it as a whole) run noticeably quicker for me locally won this branch, at e.g. the whole suite takes ~50 seconds, relative to ~65s with the current `main`.

One version reference needs updating but otherwise all good. Please merge when ready.

davidhassell:
Thanks, Sadie.

PR #264, opened 2023-06-20
Docs: remove or replace cases of some subjective filler terms

sadielbartholomew:
Equivalent to NCAS-CMS/cf-python#670. Should be quite self-explanatory but see my opening description on that issue for context if required.


davidhassell (approved):
Much improved. Thanks, Sadie.

PR #265, opened 2023-06-21
Remove archived documentation from repo, point to new location

sadielbartholomew:
Removes the built archived documentation (essentially that contained in the docs/1.* directories) which has moved to its new home in the [NCAS-CMS/cfdm-docs repository](https://github.com/NCAS-CMS/cfdm-docs) and updates the links to such archived docs so that the latest documentation points to the new location.

Note for reviewing the following:
* This PR is equivalent to  NCAS-CMS/cf-python#663 for cf-python.
* The initial link to the 'Version 1.10.1.0' archived docs will not work because that latest set of docs pages was not generated to the main 'cfdm' repo in order to be moved over to 'cfdm-docs', but all other links should work. We should work out how to add v1.10.1.0 retrospectively to the cfdm-docs repo after this is merged.


davidhassell (approved):
Hi Sadie - looks good. I haven't really checked all 5000+ file changes, but I'm sure it's fine :)

PR #267, opened 2023-08-30
Refactor code for writing netCDF unlimited dimensions

davidhassell:
Refactor code for writing netCDF unlimited dimensions: Create a wrapper method in `netcdfwrite.py` to call `cfdmimplementation.nc_is_unlimited_axis`, rather than than the latter being called directly. (The wrapper already existed, but was not called.)

This is so that subclasses can modify the behaviour if needs be. E.g., cf-python needs to never write unlimited dimension to CFA-netCDF files.


sadielbartholomew (approved):
Makes the relevant code a bit more readable and tidy. All working, still. Great, please merge.

PR #273, opened 2023-10-19
Remove Python 3.7

davidhassell:
Fixes #274

PR #275, opened 2023-10-23
Actions: Python versions update to fix `int` misinterpretation

sadielbartholomew:
Minor fix. Sorry, I missed this in review of #273, but we need to be careful when quoting versions such as `3.10` which are subject to integer simplification in the YAML parsing, so interpreted as, for example in this case, `3.1`, when not quoted. Hence we try to run on Python 3.1 in the CI jobs (see for instance the run log for https://github.com/NCAS-CMS/cfdm/actions/runs/6610035829/job/17951188032).

The safest approach to fix this, as widely used and [advised](https://github.com/actions/runner/issues/1989), is to provide versions as strings, so I've set that across the workflows (even for versions which won't be misinterpreted, for consistency and safety).

### DIFF
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -20,7 +20,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, ubuntu-18.04, macos-latest, macos-10.15]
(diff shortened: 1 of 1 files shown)


davidhassell (approved):
Good catch.

PR #277, opened 2023-10-31
Detection of duplicates in API reference, some to remove

sadielbartholomew:
Took a slight tangent from my reviewing of #272 when considering API reference coverage and spotting there are some methods listed multiple times in the same file or even under the same sub-heading.

I thought I'd make a small update to the `check_docs_api_coverage` script so we can automate the detection of those when we are already checking that they are all present. So with this change we are checking they are all present precisely once and not just not zero times like before (sub-string method name complications are dealt with - see code comments), unless we've deliberately chosen to duplicate them under different headings. A warning is raised reporting any duplicates for consideration.

I've removed any duplication that seems unintended in the API reference source restructured text, thought left in some cases where it looks like we might want to put methods under multiple headings. Namely, at PR opening the script still reports these which I haven't touched:

```console
$ ./check_docs_api_coverage
WARNING: some methods are listed multiple times inside one class file/page. Decide if the duplicates are intended and if not remove them. They are:
cfdm.Data.del_calendar
cfdm.Data.get_calendar
cfdm.Data.has_calendar
cfdm.Data.set_calendar
cfdm.Data.sum
cfdm.Data.unique
cfdm.Field.del_data_axes
cfdm.Field.get_data_axes
cfdm.Field.has_data_axes
cfdm.Field.set_data_axes
cfdm.core.Coordinate.del_bounds
cfdm.core.Coordinate.del_data
cfdm.core.Coordinate.set_bounds
cfdm.core.Coordinate.set_data
cfdm.core.Data.del_calendar
cfdm.core.Data.get_calendar
cfdm.core.Data.has_calendar
cfdm.core.Data.set_calendar
cfdm.core.Field.del_data_axes
cfdm.core.Field.get_data_axes
cfdm.core.Field.has_data_axes
cfdm.core.Field.set_data_axes
cfdm.core.PropertiesDataBounds.del_bounds
cfdm.core.PropertiesDataBounds.del_data
cfdm.core.PropertiesDataBounds.set_bounds
cfdm.core.PropertiesDataBounds.set_data

All non-private methods are documented
```

as opposed to, previously, a longer list:

```console
WARNING: there are methods which are listed multiple times in one class file/page. Decide if the duplicates are useful. They are:
cfdm.Data.del_calendar
cfdm.Data.get_calendar
cfdm.Data.has_calendar
cfdm.Data.set_calendar
cfdm.Data.sum
cfdm.Data.unique
cfdm.Domain.apply_masking
cfdm.Domain.climatological_time_axes
cfdm.Field.climatological_time_axes
cfdm.Field.del_data_axes
cfdm.Field.get_data_axes
cfdm.Field.has_data_axes
cfdm.Field.set_data_axes
cfdm.NetCDFArray.get_address
cfdm.core.Coordinate.del_bounds
cfdm.core.Coordinate.del_data
cfdm.core.Coordinate.set_bounds
cfdm.core.Coordinate.set_data
cfdm.core.Data.del_calendar
cfdm.core.Data.get_calendar
cfdm.core.Data.has_calendar
cfdm.core.Data.set_calendar
cfdm.core.DomainAxis.construct_type
cfdm.core.Field.del_data_axes
cfdm.core.Field.get_data_axes
cfdm.core.Field.has_data_axes
cfdm.core.Field.set_data_axes
cfdm.core.PropertiesDataBounds.del_bounds
cfdm.core.PropertiesDataBounds.del_data
cfdm.core.PropertiesDataBounds.set_bounds
cfdm.core.PropertiesDataBounds.set_data


All non-private methods are documented
```

Happy to remove those duplicates from the first snippet too, though? But the main thing is the automation of the check and that I've removed methods duplicated under the same heading, which can be blitzed to tidy.

### DIFF
--- a/docs/source/check_docs_api_coverage.py
+++ b/docs/source/check_docs_api_coverage.py
@@ -16,6 +16,7 @@ Call as:
 
 """
 import os
+import re
 import sys
 
 import cfdm as package
@@ -77,7 +78,30 @@ for core in ("", "_core"):
                         f"{os.path.join(source, 'class', rst_file)}"
                     )
                 elif count > 1:
-                    duplicate_method_entries.append(method)
+                    # The method appears more than once, but may be a
+                    # sub-string of another method name, e.g. this gets caught:
+                    #     [cfdm.List.]nc_set_variable
+                    # due to the presence of this method:
+                    #     [cfdm.List.]nc_set_variable_groups
+                    # so we must account for that. Checking next character
+                    # of duplicate(s) is not an underscore, is easiest.
+                    locs = [
+                        (m.start(0), m.end(0))
+                        for m in re.finditer(method, rst_contents)
+                    ]
+                    chars_at_next_loc = [
+                        m[1] for m in [lcs for lcs in locs]  # == last char + 1
+                    ]
+                    chars = [rst_contents[c] for c in chars_at_next_loc]
+
+                    newline_count = chars.count("\n")
+                    ws_count = chars.count(" ")
+                    # Any character that isn't a newline or whitespace
+                    # indicates another method which the method is a substring
+                    # of and can be excluded. If there are still duplicates,
+                    # we have genuine duplicate listing entries to report.
+                    if newline_count + ws_count > 1:
+                        duplicate_method_entries.append(method)
         except FileNotFoundError:
             n_missing_files += 1
             print(f"File {rst_file} does not exist")
--- a/docs/source/check_docs_api_coverage.py
+++ b/docs/source/check_docs_api_coverage.py
@@ -84,23 +84,18 @@ for core in ("", "_core"):
                     # due to the presence of this method:
                     #     [cfdm.List.]nc_set_variable_groups
                     # so we must account for that. Checking next character
-                    # of duplicate(s) is not an underscore, is easiest.
-                    locs = [
-                        (m.start(0), m.end(0))
-                        for m in re.finditer(method, rst_contents)
+                    # of duplicate(s) is something other than a newline or
+                    # whitespace, seems robust and simplest.
+                    end_loc = [
+                        m.end(0) for m in re.finditer(method, rst_contents)
                     ]
-                    chars_at_next_loc = [
-                        m[1] for m in [lcs for lcs in locs]  # == last char + 1
-                    ]
-                    chars = [rst_contents[c] for c in chars_at_next_loc]
+                    chars = [rst_contents[c] for c in [e for e in end_loc]]
 
-                    newline_count = chars.count("\n")
-                    ws_count = chars.count(" ")
                     # Any character that isn't a newline or whitespace
                     # indicates another method which the method is a substring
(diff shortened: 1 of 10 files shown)


davidhassell:
Hi Sadie - thanks for this. I reckon that we could keep `cfdm.Data.sum` and `cfdm.Data.unique` and get rid of the rest. What do you think?

sadielbartholomew:
> I reckon that we could keep cfdm.Data.sum and cfdm.Data.unique and get rid of the rest. What do you think?

Sounds good to me. I'll update the PR shortly to get the duplicates other than those removed, in the least appropriate location.

sadielbartholomew:
@davidhassell I have now updated the PR to remove the majority of the rest of the duplicates, so that the output of the script is:

```console
$ ./check_docs_api_coverage 
WARNING: some methods are listed multiple times inside one class file/page. Decide if the duplicates are intended and if not remove them. They are:
cfdm.Data.sum
cfdm.Data.unique

All non-private methods are documented
```

in line with your idea to leave just those two under `cfdm.Data`.

So this should be ready for re-review. Thanks.

docs/source/check_docs_api_coverage.py
```diff
@@ -68,12 +70,33 @@
...
+                elif count > 1:
+                    # The method appears more than once, but may be a
+                    # sub-string of another method name, e.g. this gets caught:
+                    #     [cfdm.List.]nc_set_variable
+                    # due to the presence of this method:
+                    #     [cfdm.List.]nc_set_variable_groups
+                    # so we must account for that. Checking next character
+                    # of duplicate(s) is something other than a newline or
+                    # whitespace, seems robust and simplest.
+                    end_loc = [
```
davidhassell, line 89: I presume I'm missing something here, but `count` doesn't count partial matches, so why is this needed?
```python
>>> a = ['abc', 'abcd', 'abcde']
>>> a.count('abc')
1

**Review by davidhassell — COMMENTED** [review]
Hi Sadie, looks good, thanks. I have a question about the partial matches, though ...

File: docs/source/check_docs_api_coverage.py (hunk 1 above)
Comment (sadielbartholomew) [reply docs/source/check_docs_api_coverage.py:89]: I think it's needed because what's being done is count from one larger string, being the string read from the full file, rather than a list of sub-strings as per your example, in which case we have as an equivalent:

```python
>>> a = "abc abcd abcde"
>>> a.count("abc")
3

```

sadielbartholomew: Regardless, if you can think of a more elegant way to go about the logic I've added, I am happy to hear. I might be able to think of an alternative myself if I mull over it, but since it's a convenience script for automating checks I am satisfied as long as it works and isn't (too) slow :)

davidhassell: All good - thanks for the explanation

davidhassell (approved):
Following the short discussion, I now understand, and therefore approve, all of these changes. Thanks, Sadie.

PR #283, opened 2024-01-05
Change `numpy` to `np`

### LINKED ISSUE #278: Always import `numpy` as `np`
For consistency, we should replace all `import numpy` with `import numpy as np`, and update all occurrences of `numpy.` with `np.`.

davidhassell:
Fixes #278 

(No rush on this one!)

### DIFF
--- a/cfdm/cellmethod.py
+++ b/cfdm/cellmethod.py
@@ -1,7 +1,7 @@
 import logging
 from copy import deepcopy
 
-import numpy
+import numpy as np
 
 from . import core, mixin
 from .data import Data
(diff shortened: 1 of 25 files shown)


sadielbartholomew (approved):
Brill, thanks. As for closing #278, there are still some `import numpy` uses in the test suite, based on a `git grep "import numpy"` and eyeball to see if any lacked the `as np` at the end at this stage, but it seems you've caught all of the ones outside of it. But maybe we can change those too for this PR, whilst we are at it (though there's an argument to make those all, consistently, `import numpy` as opposed to the other way round given it is more explicit for testing and we don't seem to use many, if any, `as` aliases in the test suite at present)?

Either way, happy for you to merge this, though preferably we address the inconsistency in the `import`s int he test suite too first.

davidhassell:
Thanks Sadie. Test suite now updated to `import numpy as np` (82634e31b077b71a483e6d902e1e63e1fad2d3bd).

sadielbartholomew (approved):
Thanks @davidhassell for updating this. I've sanity-checked the extra commit and it's all fine, so please merge.

PR #284, opened 2024-01-05
CI: fixes and updates to pre-commit workflow for linting pass

sadielbartholomew:
#283 shows that the linting workflow is running into numerous failures, mostly E231, which I don't see locally. This is to update the workflow and/or make fixes so we are up-to-date and can see only actual linting issues.

[WIP as I check the results of the CI jobs, since locally the linting passes for me so it us harder to isolate there what versions etc. are causing the (majority of the) error code violation raises.]

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,7 +59,7 @@ repos:
   # (see https://flake8.pycqa.org/en/latest/ for documentation and see
   # the cfdm .flake8 file for our custom flake8 configuration)
   - repo: https://github.com/PyCQA/flake8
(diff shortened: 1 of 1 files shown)


sadielbartholomew:
Oh, excellent - I was thinking it might be more complicated, hence the PR rather than lone commit(s) to address, but all that was needed was (my first guess) to update `flake8` from the older (late 2022 release) 6.0.0 to 6.1.0 ([yesterday](https://github.com/PyCQA/flake8/tags) the first 7. version was released but since it might be unstable I'll wait a little before updating us to that).

Trivial PR so merging.

PR #296, opened 2024-05-17
`netcdfread`: safer version handling for `Conventions` property

### LINKED ISSUE #295: Error reading files with pattern-defying CF Conventions value format
I recently encountered a file (from investigating spammy warnings which brought me to https://github.com/Unidata/cftime/issues/328 - see the file attached in the opening comment there as an example) which had the `Conventions` global attribute value of `:Conventions = "CF-1.6/CF-1.7"` (checked via `ncdump -h`), , a compound form which isn't standard that cfdm can't read it because it errors on processing the version in a naive way, taking whatever it finds after matching the first "CF-" pattern if found:

```python
>>> import cfdm
/home/slb93/git-repos/cfdm/cfdm/read_write/netcdf/netcdfread.py:1028: SyntaxWarning: invalid escape sequence '\s'
  all_conventions = re.split(",\s*", Conventions)
>>> cfdm.read("~/Downloads/subset.nc")
  File "<stdin>", line 1, in <module>
  File "/home/slb93/git-repos/cfdm/cfdm/read_write/read.py", line 328, in read
    fields = netcdf.read(
             ^^^^^^^^^^^^
  File "/home/slb93/git-repos/cfdm/cfdm/decorators.py", line 171, in verbose_override_wrapper
    return method_with_verbose_kwarg(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/slb93/git-repos/cfdm/cfdm/read_write/netcdf/netcdfread.py", line 1056, in read
    g["file_version"] = Version(file_version)
                        ^^^^^^^^^^^^^^^^^^^^^
  File "/home/slb93/miniconda3/envs/cf-env-312/lib/python3.12/site-packages/packaging/version.py", line 200, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: '1.6/CF-1.7'
```

Whilst a weird-y value such as 'CF-1.6/CF-1.7' is not the CF-compliant value to set on that attribute and we shouldn't account for any weirdness that data may possess, IMO it shouldn't mean such files can't be read in at all. I looked at the logic of the 'Conventions' property processing and concluded that it isn't very robust and should be improved so that weird edge cases don't error and instead any non-standard and therefore ambiguous values such as this are ignored - the files can be read in but the CF version is considered ambiguous therefore gets set by our default logic for lack of known/set version.

PR to follow, which makes the `Conventions` attribute versions processing more robust through regular expressions.

### LINKED ISSUE #297: Error on reading files with conformant Conventions value 'CF-1.*-draft'
The working CF Conformance document. i.e. the one in draft form, states that (using the example of the current working document i.e. v.1.12)

> Files that conform to the CF version 1.12 draft conventions must indicate this by setting the global Conventions attribute to contain the CF string value "CF-1.12-draft".

This clearly gets changed to 'CF-1.12' when the version is released such that it is no longer draft form, but to me that implies that "CF-1.12-draft" and similarly any 'CF-\<valid version string\>-draft' value should strictly be a conformant value for the Conventions global attribute. However, when we use cfdm (current main branch, and same for 1.11.1.0/last release) to read in such a file, it errors:

```python
>>> import cfdm
>>> import netCDF4
>>> f = cfdm.example_field(0)
>>> cfdm.write(f, "delme.nc")
>>> n = netCDF4.Dataset("delme.nc", "a")
>>> n.Conventions
'CF-1.11'
>>> n.Conventions = "CF-1.12-draft"
>>> n.Conventions
'CF-1.12-draft'
>>> n.close()
>>> g = cfdm.read("delme.nc")[0]
  File "<stdin>", line 1, in <module>
  File "/home/slb93/git-repos/cfdm/cfdm/read_write/read.py", line 328, in read
    fields = netcdf.read(
             ^^^^^^^^^^^^
  File "/home/slb93/git-repos/cfdm/cfdm/decorators.py", line 171, in verbose_override_wrapper
    return method_with_verbose_kwarg(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/slb93/git-repos/cfdm/cfdm/read_write/netcdf/netcdfread.py", line 1056, in read
    g["file_version"] = Version(file_version)
                        ^^^^^^^^^^^^^^^^^^^^^
  File "/home/slb93/miniconda3/lib/python3.11/site-packages/packaging/version.py", line 198, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: '1.12-draft'
```

Whereas such a file should be able to be read in without issue. This can be addressed as part of my PR #296.

### LINKED ISSUE #298: NetCDF reading `default_version` argument is not exposed to user: remove it?
The netCDF-specific `read`, i.e. `NetCDFRead.read` method, has an argument `default_version` which isn't documented but according to a comment over logic using the corresponding variable, it supports a "default version provided by the user" for the Conventions:

https://github.com/NCAS-CMS/cfdm/blob/e948eef4abf65b12fe1c0f8785f185ed9a0c3cd0/cfdm/read_write/netcdf/netcdfread.py#L774-L778

however in our user-exposed `read` function, there is no way to access it:

https://github.com/NCAS-CMS/cfdm/blob/e948eef4abf65b12fe1c0f8785f185ed9a0c3cd0/cfdm/read_write/read.py#L11-L21

so there is no way to apply that in the intended way, as a user. It is not documented at all across the documentation either.

_It seems like this is dead code which needs to be removed, but I've opened this as a question in case we instead want to revive it by exposing it through the `read` function._ @davidhassell, what do you think is best?

If we want to remove the several lines of logic that concern it, I can do so as part of #296 which touches the same area of code.

### LINKED ISSUE #301: Prepare for UGRID versions greater than 1.0
The code presently registers and stores any valid version ID (1.0, 2.10.4, 6.2.3, etc.) for the UGRID Conventions, but only processes any UGRID features with code supporting v1.0. This is probably harmless for now with v1.0 being the latest and greatest version (and that included in CF for the present CF, v1.11), but as-is the behaviour will be mis-leading for subsquent UGRID versions upon release. To be ready in case datasets are around ready/compliant for new UGRID releases, we should at least warn that the code is assuming and processing features according to UGRID v1.0.

As discussed offline. To be fixed in #296.

sadielbartholomew:
Close #295, ~~close #297,~~ close #298, close #301.

~~Waiting on some clarifications to https://github.com/orgs/cf-convention/discussions/321 and to https://github.com/NCAS-CMS/cfdm/issues/298 before updating and converting from draft form ready for review.~~
https://github.com/orgs/cf-convention/discussions/321 has confirmed that we don't have to worry about a '-draft' possibility in the `Conventions` attribute value, so #297 is a non-issue I have closed. I will now update the PR accordingly to revert anything related to that and open it for review.

### DIFF
--- a/cfdm/read_write/netcdf/netcdfread.py
+++ b/cfdm/read_write/netcdf/netcdfread.py
@@ -1030,23 +1030,25 @@ class NetCDFRead(IORead):
             all_conventions = Conventions.split()
 
         file_version = None
-        valid_version_pattern = re.compile("(\d+.\d+)")
         for c in all_conventions:
-            cf_v = re.findall(r"CF-(.*)", c)
-            u_v = re.findall(r"UGRID-(.*)", c)
-            cfa_v = re.findall(r"CFA-(.*)", c)
+            # Be particularly strict with the regex to account for ambiguous
+            # values e.g. CF-<badversionformat> or CF-1.X/CF-1.Y. Note that
+            # the '^' and '$' start and end of string tokens ensure that
+            # only zero or one match can be found per given string c (hence
+            # taking group(1) when given conditional is True below is safe).
+            cf_v = re.search(r"^CF-(\d+.\d+)$", c)
+            u_v = re.search(r"^UGRID-(\d+.\d+)$", c)
+            cfa_v = re.search(r"^CFA-(\d+.\d+)$", c)
 
-            # Else ambiguous e.g. CF-<badversionformat> or CF-1.X/CF-1.Y
-            if len(cf_v) == 1 and re.findall(valid_version_pattern, cf_v[0]):
-                file_version = cf_v[0]
-            elif len(u_v) == 1 and re.findall(valid_version_pattern, u_v[0]):
+            if cf_v:
+                file_version = cf_v.group(1)
+            elif u_v:
                 # Allow UGRID if it has been specified in Conventions,
                 # regardless of the version of CF.
-                g["UGRID_version"] = Version(u_v[0])
-            elif len(cfa_v) == 1 and re.findall(
-                    valid_version_pattern, cfa_v[0]):
+                g["UGRID_version"] = Version(u_v.group(1))
+            elif cfa_v:
                 g["cfa"] = True
-                g["CFA_version"] = Version(cfa_v[0])
+                g["CFA_version"] = Version(cfa_v.group(1))
             elif c == "CFA":
                 g["cfa"] = True
                 g["CFA_version"] = Version("0.4")
--- a/cfdm/read_write/netcdf/netcdfread.py
+++ b/cfdm/read_write/netcdf/netcdfread.py
@@ -1020,9 +1020,10 @@ class NetCDFRead(IORead):
 
         # ------------------------------------------------------------
         # Find the CF version for the file, and the CFA version.
+        # (See '2.6.1 Identification of Conventions' in the CF Conformance
+        # document for valid inputs for the 'Conventions' property.)
         # ------------------------------------------------------------
         Conventions = g["global_attributes"].get("Conventions", "")
-
         # If the string contains any commas, it is assumed to be a
         # comma-separated list.
         all_conventions = re.split(r",\s*", Conventions)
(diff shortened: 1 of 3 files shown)


sadielbartholomew:
Closing due to being stale - I will shortly re-review the issues referenced and whether they are still relevant and need attention.

PR #300, opened 2024-05-30
Fix bug that returned incorrect results when an invalid identifer is provided to `cf.Field.cell_methods`

### LINKED ISSUE #299: `cf.Field.cell_methods` returns incorrect results when an invalid identifer is provided
At cfdm `v1.11.1.0`, when an invalid selection is passed to `cfdm.Field.cell_methods`, _all_ cell method constructs are returned, rather than none of them as expected:
```python
>>> import cfdm
>>> f = cfdm.example_field(1)
>>> print(f.cell_methods())  # CORRECT output:
Constructs:
{'cellmethod0': <CellMethod: domainaxis1: domainaxis2: mean where land (interval: 0.1 degrees)>,
 'cellmethod1': <CellMethod: domainaxis3: maximum>}
>>> print(f.cell_methods('bad identifier'))  # WRONG output, expected: Constructs: {}
Constructs:
{'cellmethod0': <CellMethod: domainaxis1: domainaxis2: mean where land (interval: 0.1 degrees)>,
 'cellmethod1': <CellMethod: domainaxis3: maximum>}
```

davidhassell:
Fixes #299


sadielbartholomew (approved):
One typo in the changelog and formatting issue, but otherwise perfect with sufficient new testing. +1 for the new examples for the docs. Please merge once the minor comments are addressed.

(FYI as part of testing, in case of logic re-use and because bugs can,  I did a quick check that other multiple construct getting methods don't display the same bug, e.g. `f.coordinate_references` and `f.domain_axes('bad')`, but all looks good.)

PR #326, opened 2025-03-03
Allow dask>2024.7.1 

### LINKED ISSUE #321: Allow `dask>2024.7.1`
Currently, Dask is pinned: `dask<=2024.7.1`, due to Dask API changes after that version.

We should allow newer versions of Dask, creating a new minimum version (e.g. 2025.1.0), rather than insisting a maximum.

 The issues for cfdm with newer versions of Dask are:

1. The new Dask "task spec" graph API breaks the current cfdm functionality that inspects graphs for `Array` objects (e.g.  `cfdm.Data._modify_dask_graph`).
2. The new Dask `take` function sometimes requires "an implementation of 'numpy.concatenate' on types that implement \_\_array_function__", i.e. `Array` objects - which don't currently have this.

PR to follow to fix these items

davidhassell:
Fixes #321


cfdm/test/test_Domain.py
```diff
@@ -227,12 +227,13 @@ def test_Domain_persist(self):
...
         cfdm.write(f, tmpfile)
-        f = cfdm.read(tmpfile)[0].domain
+        f = cfdm.read(tmpfile, dask_chunks=3)[0][:2, :2].domain
 
         on_disk = False
-        for v in f.coordinate("longitude").data.todict().values():
+        for k, v in f.coordinate("longitude").data.todict().items():
             if isinstance(v, cfdm.data.abstract.FileArray):
                 on_disk = True
+                break
```
sadielbartholomew, line 236: Not needed? I guess this was put in for dev work and not removed?

```suggestion
```

cfdm/test/test_Domain.py
```diff
@@ -241,6 +242,7 @@ def test_Domain_persist(self):
         for v in g.coordinate("longitude").data.todict().values():
             if isinstance(v, np.ndarray):
                 in_memory = True
+                break
```
sadielbartholomew, line 245:
```suggestion
```

cfdm/test/test_Domain.py
```diff
@@ -251,6 +253,7 @@ def test_Domain_persist(self):
         for v in f.coordinate("longitude").data.todict().values():
             if isinstance(v, np.ndarray):
                 in_memory = True
+                break
```
sadielbartholomew, line 256:
```suggestion
```

sadielbartholomew (approved):
(Reviewing in tandem with the cf-python PR, so will submit my review for that in a moment.) I am happy with this branch (i.e. on the cfdm side), with a few minor comments on the code changes. And functionally, all is good with sufficient new testing - works here and working downstream with the cf-python branch too, hence approving on PR https://github.com/NCAS-CMS/cf-python/issues/849 also.

As discussed this morning in the video call, I am not testing on performance since I am assuming this will not significantly affect the performance of either cfdm or cf-python. So as long as you are happy to assume that, too (else I can review this in that respect, too).

docs/source/installation.rst
```diff
@@ -183,6 +210,28 @@ The cfdm package requires:
 * `cfunits <https://pypi.org/project/cfunits>`_, version 3.3.7 or
   newer.
 
+.. _UNIDATA-UDUNITS-2-library:
+
+* `UNIDATA Udunits-2 library
+  <http://www.unidata.ucar.edu/software/udunits>`_, version 2.2.25
+  or newer. UDUNITS-2 is a C library that provides support for units of
```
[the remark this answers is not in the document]
davidhassell: I think that UDUNITS is actually the correct format (https://www.unidata.ucar.edu/software/udunits, and as used in CF) - so I'll change other things to capitals.

cfdm/test/test_Domain.py
davidhassell: It's OK isn't it? No need to iterate through more of the loop than needed?

docs/source/installation.rst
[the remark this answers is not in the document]
davidhassell: 3b031fabedac3c5304c5343ca3b19acaaa943585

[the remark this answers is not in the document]
sadielbartholomew: Excellent, thanks - all caps was in fact my preferred form, so that's ideal. Marking as resolved.

cfdm/test/test_Domain.py
sadielbartholomew: Ah yes of course, sorry, I clearly didn't look carefully enough and though it was an indent level down and skipped everything. My bad. Marking all as resolved.

PR #333, opened 2025-03-31
Add `astype` method on `Array` subclasses. Pin Dask.

### LINKED ISSUE #331: Missing `astype` method on `Array` subclasses
As described in https://github.com/NCAS-CMS/cf-python/issues/856, dask at versions 2025.2.0 or 2025.3.0 (at least) sometimes has a problem because subclasses of `Array` (e.g. `NetCDF4Array` and `H5netcdfarray`) do not have an `astype` method, which can cause the following exception from Dask:

```python
File ~/miniconda3/lib/python3.12/site-packages/dask/array/chunk.py:279, in astype(x, astype_dtype, **kwargs)
    278 def astype(x, astype_dtype=None, **kwargs):
--> 279     return x.astype(astype_dtype, **kwargs)

AttributeError: 'NetCDF4Array' object has no attribute 'astype'
```

It's not clear to me exactly why, but I think it's like coming out of the slicing process, in certain circumstances.

The solution is simply to give these cfdm classes an `astype` method that returns the appropriate numpy array.

### LINKED ISSUE #332: Pin dask to `dask>=2025.2.0,<=2025.3.0`
Dask is on the move again - I notice that `dask.base.collections_to_dsk` has gone from the Dask main branch, _after_ the release of Dask 2025.3.0.

It's a real breaker, because it stops `cfdm.Data.todict` from working, so best pin to `dask>=2025.2.0,<=2025.3.0` for now.

I'll open another issue for fixing things up for `dask>2025.3.0` - it might be quite an easy fix (using the new `dask.base.collections_to_expr`, perhaps?), but we should wait until that's been publicly released.

davidhassell:
Fixes #331
Fixes #332


sadielbartholomew (approved):
All good, with the bug MRE now no longer erroring and sufficient new testing added. Thanks for doing the fix. Please merge.

PR #334, opened 2025-04-07
Cache cached data elements in `cfdm.read`

### LINKED ISSUE #313: Improve performance in `cfdm.read` by caching any array values retrieved from disk
First and last array values are cached in `NetCDFRead.read._cache_data_elements`, and it currently could do this multiple times per variable.

It would be straight forward to cache these values to save going back to the dataset more often than is required. 

PR to follow.

davidhassell:
Fixes #313


cfdm/read_write/netcdf/netcdfread.py
```diff
@@ -11273,6 +11288,9 @@ def _cache_data_elements(self, data, ncvar):
 
             elements[index] = value
 
+        # Cache the cached data elements for this variable
+        g["cached_data_elements"][ncvar] = elements
```
sadielbartholomew, line 11292: Why do we run also run `data._set_cached_elements(elements)` in the line after this? It that (still) necessary?

sadielbartholomew (approved):
One question, but overall perfect so please merge once answered. Thanks.

cfdm/read_write/netcdf/netcdfread.py
davidhassell: Hi Sadie - `data._set_cached_elements(elements)` caches the elements on the `Data` object, but the way the code is at the moment, all metadata netCDF variables get read twice (once as a potential field, and once as metadata), so we want to cache the elements at the `netcdfread.py` level, too, so that the second read doesn't have to go back to the file.

davidhassell: Oh, hang on! But elements aren't cached for Field constructs, so there should be no double-reading at all ...  In that case maybe this issue is not an issue at all! I'll have a think and a dig ....

PR #344, opened 2025-07-21
Add `distributed` as required dependency with version pin

sadielbartholomew:
As per NCAS-CMS/cf-python#876. Doing here also for safety. The two PRs are pretty much mirrors of each other, except with respect to the `cf[dm].environment` functions notably where `cfdm` handles the report on `distributed` for cf-python to spit out as part of the cfdm requirements.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,12 @@
+Version NEXTVERSION
+--------------
+
+**2025-??-??**
+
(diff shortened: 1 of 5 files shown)


sadielbartholomew:
Linting CI check failing due to some pre-commit issue ("Our services aren't available right now") so we can ignore that. Merging...

PR #346, opened 2025-07-25
Provide option to `Field.dump` to not display the first and last data values

### LINKED ISSUE #345: Provide option to `Field.dump` to not display the first and last data values
Displaying the first and last Field data values in `cfdm.Field.dump` can take a long time if the data need an expensive computation (possibly including a slow read from local or remote disk). It would be nice to have an option to skip this display of values in these cases, so that `dump` doesn't take too long.
```python
>>> f.dump()
<snip>
Data(latitude(5), longitude(8)) = [[0.007, ..., 0.013]] 1
<snip>

>>> f.dump(data=False)
<snip>
Data(latitude(5), longitude(8))
<snip>
```

PR to follow.

davidhassell:
Fixes #345

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -5,6 +5,7 @@ Version NEXTVERSION
 
 * New keyword parameter to `cfdm.Field.dump`: ``data``
   (https://github.com/NCAS-CMS/cfdm/issues/345)
+* New dependency: ``distributed>=2025.5.1``
 
 ----
 
--- a/cfdm/__init__.py
+++ b/cfdm/__init__.py
@@ -166,6 +166,20 @@ else:
             f"Got {dask.__version__} at {dask.__file__}"
         )
 
+# Check the version of distributed
+try:
+    import distributed
+except ImportError as error1:
+    raise ImportError(_error0 + str(error1))
+else:
+    _minimum_vn = "2025.5.1"
+    if Version(distributed.__version__) < Version(_minimum_vn):
+        raise ValueError(
+            "Bad distributed version: cfdm requires "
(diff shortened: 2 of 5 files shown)


sadielbartholomew (approved):
Good idea. It works and everything looks good implementation wise.

My main comment is that we might want to include the units in the data line, given for `print(f)` and `f` (i.e. `str` and `repr`) they don't show the data but do indicate the units, however one could argue the units are shown in the global properties so that's sufficient. Will leave it up to you.


```python
>>> f
<Field: air_temperature(atmosphere_hybrid_height_coordinate(1), grid_latitude(10), grid_longitude(9)) K>
>>> print(f)
Field: air_temperature (ncvar%ta)
---------------------------------
Data            : air_temperature(atmosphere_hybrid_height_coordinate(1), grid_latitude(10), grid_longitude(9)) K
...
>>> f.dump(display=True, data=False)
---------------------------------
Field: air_temperature (ncvar%ta)
---------------------------------
Conventions = 'CF-1.12'
project = 'research'
standard_name = 'air_temperature'
units = 'K'

Data(atmosphere_hybrid_height_coordinate(1), grid_latitude(10), grid_longitude(9)) [add unit symbol here e.g. K as per the above reports?]
...
```

Also, ideally you can update the second-line docstring description for `dump` so that "... and provides selected values of all data arrays." is qualified with something like "... unless requested otherwise"/"optionally excluding the variable data" (at the end).

Please merge when considered.

davidhassell:
Hi Sadie,

> we might want to include the units in the data line

That's a good idea: e42e09cd35b88740ddc178e58a7e4aae430af6c8

PR #349, opened 2025-08-06
Fix `cfdm.Data.reshape` when the underlying data originate on disk

### LINKED ISSUE #348: `reshape` fails when the underlying data originate on disk
At cfdm v1.12.2.0, `cfdm.Data.reshape` fails when the underlaying data in the Dask array is a `FileAarray` object, as opposed to an array in memory. I.e. when the the data is being pointed to in a file on disk, the `reshape` operation fails:
```python
>>> import cfdm
>>> f = cfdm.example_field(0)
>>> f = cfdm.write(f, 'test.nc')
>>> f = cfdm.read('test.nc')[0]
>>> d = f.data
>>> d.shape
(5, 8)
>>> d = d.reshape(8, 5)
>>> d.array
AttributeError: 'H5netcdfArray' object has no attribute 'reshape'
```
This is due a beneficial refactor of `dask.array.shape`, which now no longer relies on `__getitem__` to work, and so can no longer work on `FileArray` objects (like `H5netcdfArray`).

The solution is simply to make sure that the data are read into memory (only at compute time!) before the reshape operation is added to the Dask graph - a very small change to implement. PR to follow.

Thanks to @CheeseFace for pointing this out.

davidhassell:
Fixes #348

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -3,6 +3,8 @@ Version NEXTVERSION
 
 **2025-??-??**
 
+* Fix `cfdm.Data.reshape` when the underlying data originate on disk
+  (https://github.com/NCAS-CMS/cfdm/issues/348)
 * New keyword parameter to `cfdm.Field.dump`: ``data``
   (https://github.com/NCAS-CMS/cfdm/issues/345)
 * New dependency: ``distributed>=2025.5.1``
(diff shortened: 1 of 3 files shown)


sadielbartholomew (approved):
Perfect - with a nice concise test to isolate the now-fixed bug. Please merge.

PR #357, opened 2025-10-01
Allow `#` in file names

### LINKED ISSUE #356: `cfdm.read` fails to read a file with `#` characters in the file name
In versions `1.12.2.0` and `1.12.3.0` a bug was introduced that a file from being read if it contained the `#` character. 

This is a result of the internal use of  `uritools.urisplit` to parse the file name, which (reasonably) treats `#` as the separator for a [URI fragment]( https://datatracker.ietf.org/doc/html/rfc3986#section-3.5), and was, with our usage of `urisplit`, stripping off the fragment, resulting in a file name that didn't exist.

For instance, `xpcvba#pa000007543c1+` was being mangled to `xpcvba`, resulting in a `FileNotFoundError` exception coming from `cfdm.read`.

PR to follow.

Thanks to Charlotte Lang for spotting this.

davidhassell:
Fixes #356

Essentially targets uses of `uritools.urisplit` and `uritools.uricompose` to account for URI fragments. Makes use of `cfdm.abspath` and `cfdm.dirname` where possible, as these now deal with it appropriately.

### DIFF
--- a/cfdm/test/test_read_write.py
+++ b/cfdm/test/test_read_write.py
@@ -23,6 +23,8 @@ tmpfiles = [
     tempfile.mkstemp("_test_read_write.nc", dir=os.getcwd())[1]
     for i in range(n_tmpfiles)
 ]
+tmpfiles.append(tempfile.mkstemp("#test_read_write.nc", dir=os.getcwd())[1])
+
 [
     tmpfile,
     tmpfileh,
@@ -33,6 +35,7 @@ tmpfiles = [
     tmpfilec3,
     tmpfile0,
     tmpfile1,
+    tmpfile_hash,
 ] = tmpfiles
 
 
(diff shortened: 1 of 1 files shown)


sadielbartholomew (commented):
Sensible fix that works well from local testing, but I would like a dedicated test added to cover it - should be as simple as reading in an existing field (list), writing it out with a name with 2 or 3 hash characters in and then checking it re-reads - unless there's some reason that complicates like temp files not liking the '#' in the name?

(I'm guessing having a '#' in the name of a dataset does happen in practice - else I'd be tempted to argue the users should rename such files to have standard characters only rather than us fixing a weird edge case that fails for good reason... 😃 )

davidhassell:
Test added: c5cb7472d846c842d43d069edc156beca13d473b

sadielbartholomew (approved):
Thanks for adding the test. It seems sufficiently robust, so all good now - please merge.

PR #362, opened 2025-10-22
Reduce the time taken to import `cfdm`

### LINKED ISSUE #361: Reduce the time taken to import `cfdm`
Importing `cfdm` is quite slow ... on my local computer it takes 1.4 seconds.

Another system I have tried (JASMIN) can take between 6 and 10 seconds!

Having had a look at this, there seem to be two main culprits for the slow import:
- Doc string rewriting
  - At import time, every docstring (of which there are currently 5569) is inspected for doc string substitutions, replacing any that are found
- Importing external modules that themselves have a slow import
  - The main problems here are `dask`, `scipy`, `s3fs`, `zarr`, `h5netcdf`, `uritools`, `netCDF4`

These can be improved on:
- Doc string rewriting
  - Only apply substitutions when necessary, rather than trying every possible substitution for every doc string. Only 3009 of the doc strings need rewriting, and each one of those only utilises a small number of the 110 possible substitutions.
- Importing external modules that themselves have a slow import
  - Move the `dask`, `scipy`, `s3fs`, `zarr`, `h5netcdf`, `uritools` to run time, rather than import time. Many will not ever get imported, and when they do, the time is usually negligible compared to the operation being run.

#### Results

By applying these two changes, my local import time reduces to 0.2 seconds (from 1.4 seconds - a factor of 7 speed-up). On the other system I tried, the time reduces to between 1.5 and 2.5 seconds (from between 6 and 10 seconds).

These are good enough improvements for a PR, I think ...

davidhassell:
Fixes #361 

Lots of files touched, but most of them are just moving the odd import from the module level to within a method.

Three main areas:

#### 1. Doc string rewriting

- Only apply substitutions when necessary, rather than trying every possible substitution for every doc string. Only 3009 of the doc strings need rewriting, and each one of those only utilises a small number of the 110 possible substitutions.
- ~20% of the speed-up.

#### 2. Importing external modules that themselves have a slow import

- Move the dask, scipy, s3fs, zarr, h5netcdf, uritools to run time, rather than import time. Many will not ever get imported, and when they do, the time is usually negligible compared to the operation being run.
- ~80% of the speed-up.

#### 3. Refactor CONSTANTS

As a consequence of 2., we can't initialise `chunksize` at import time (it needs `dask`). This prompted a bit of a refactor of `cfdm.constants`, `cfdm.functions.ConstantAccess`, and `cfdm.configuration`. Essentially, the original `CONSTANTS` dictionary is now removed, and replaced with the dictionary `cfdm.functions.ConstantAccess._constants`. This dictionary starts off empty, and gets populated as and  when configuration parameters are accessed. In particular, when  `cfdm.configuration` is called, it makes a special effort to full populate the dictionary (there could be a way of doing this without the pre-populating, but I couldn't easily find one that didn't play havoc with the setting of constants in a context manager.)

### DIFF
--- a/cfdm/functions.py
+++ b/cfdm/functions.py
@@ -1561,7 +1561,7 @@ class ConstantAccess(metaclass=DocstringRewriteMeta):
 
     @classmethod
     def constants(cls, copy=True):
-        """TODO."""
+        """See docstring to `ConstantAccess`."""
         out = cls._constants
         if copy:
             out = out.copy()


davidhassell:
missed a few imports :) 82b6cb266d964eb25a735a0aa0ddb21e0273ded2

A note in the thread lock in `cfdm.data.locks.py` - we don't need it to serializable any more, since we don't pass it in `dask.arra.from_array` any more, so we can use the Python builtin. This change alone sped up the import by 33% from the previous commit :)

davidhassell:
...  and a few more: 9909c2fc1a87900eb58217636435aa9f37b6b231

cfdm/data/__init__.py
```diff
@@ -1,4 +1,7 @@
-from .abstract import Array, CompressedArray, MeshArray, RaggedArray
+from .abstract import Array
+from .abstract import CompressedArray
+from .abstract import MeshArray
+from .abstract import RaggedArray
```
sadielbartholomew: For my understanding, is there some (import speed or otherwise) benefit of importing individually in this way rather than as a one-liner?

cfdm/data/netcdfindexer.py
```diff
@@ -275,6 +272,8 @@ def __getitem__(self, index):
         elif data.dtype.kind in "OSU":
             kind = data.dtype.kind
             if kind == "S":
+                from netCDF4 import chartostring  # , default_fillvals
```
sadielbartholomew:
```suggestion
                from netCDF4 import chartostring
```

cfdm/functions.py
```diff
@@ -1465,44 +1470,52 @@ def copy(self):
 
 
 class ConstantAccess(metaclass=DocstringRewriteMeta):
-    '''Base class to act as a function accessing package-wide constants.
+    """Base class to act as a function accessing package-wide constants.
 
     Subclasses must implement or inherit a method called `_parse` as
-    follows:
+    follows::
```
sadielbartholomew:
```suggestion
    follows:
```

cfdm/data/aggregatedarray.py
```diff
@@ -2,14 +2,15 @@
...
-from uritools import isuri, uricompose
 
 from ..functions import dirname
 from . import abstract
 from .fragment import FragmentFileArray, FragmentUniqueValueArray
 from .netcdfindexer import netcdf_indexer
 from .utils import chunk_locations, chunk_positions
 
+# from uritools import isuri, uricompose
+
```
sadielbartholomew: Left over from dev?

```suggestion
```

(plus extra line also introduced - suggestion won't let me include that still 😬 )

sadielbartholomew (approved):
Very hot off the press, a really nice thing to note first of all is that, by Python 3.15 (though we'd probably need to wait until that was our minimum to avoid lots of conditional logic on imports, so many years...) thanks to ['PEP 810 – Explicit lazy imports'](https://peps.python.org/pep-0810/) which was [officially accepted yesterday](https://discuss.python.org/t/pep-810-explicit-lazy-imports/104131/466) there will be no need to move any heavier imports to first run-time use, but instead they can be imported at the top-level with use of a new `lazy` keyword, e.g. `lazy from scipy.sparse import issparse`. So by then we can move all of the shifted imports back home 😃 

In the meantime - and concerning the other aspect, our rather niche dostring rewriting - great PR. Does what it says on the tin, with import time reduced on my system from:

```console
$ git checkout main  # before
$ python -X importtime -c "import cfdm"
...
...
import time:      4581 |    1130718 | cfdm
$ git checkout david/docstring-rewrite-speed # after
$ python -X importtime -c "import cfdm"
...
...
import time:      2022 |     291805 | cfdm
```

so by my calculation `291805/1130718 * 100 = 25.807053571270643` % of the previous import time - nice!  (Looks like imports are slower on my system than yours, at least with the wind blowing in whatever direction it was at the time, but it's a similar pattern of improvement comparing `main` to this branch. Not sure there's much point doing multiple tries since there may be some caching effects reducing the time taken for any further attempts.) Tests still pass and functionality unaffected as far as I can otherwise tell.

Only minor comments raised, except regarding the removal of the version constraint checks in `__init.py` - see in-line comment. Once you've considered those, please merge!

cfdm/data/__init__.py
davidhassell: I don't know why these are on different lines - I'll put them back on one line

davidhassell:
Thanks for the review, Sadie.

Issue #355, opened 2025-09-03
Read grouped Zarr datasets


davidhassell:
It would be very nice to be able to read, with `cfdm.read`, [Zarr](https://zarr-specs.readthedocs.io) datasets that contain a group hierarchy.

#335 implemented the reading of Zarr datasets, but only those with no group structure (i.e. only the root group).

Zarr v2 (xarray-style) and Zarr v3 will be supported.

PR #368, opened 2025-11-13
Zarr write + Zarr groups

### LINKED ISSUE #354: Write Zarr datasets
It would be very nice to be able to write [Zarr](https://zarr-specs.readthedocs.io) datasets with `cfdm.write`

Only Zarr v3 will be supported.

The writing of group hierarchies will be supported.

See also #335 (Read Zarr datasets) and #355 (Read grouped Zarr datasets).

davidhassell:
Fixes #354 and #355 
#### cfdm/data/aggregatedarray.py
- Change to allow URIs to be Python strngs or scalar numpy string types.
#### cfdm/data/data.py
- Include Zarr dataset sharding methods
- Include Zarr fragment class
- Move the test for an "empty" slice to the parent construct (in cfdm/mixin/propertiesdata.py), because sometime we want to allow slices on abstract data (just like numpy does)
#### cfdm/data/fragment/fragmentfilearray.py
- Include Zarr fragment class
#### cfdm/data/fragment/fragmentzarrarray.py
- New file defining a Zarr fragment (very similary to `cfdm/data/fragment/fragmentnetcdf4array.py`
#### cfdm/data/netcdfindexer.py
- Account for `zarr` using the new numpy `T` data type
#### cfdm/mixin/netcdf.py
- New mixin class for dataset shards
#### cfdm/read_write/abstract/abstractio.py
- Rename "file" to "dataset"
#### cfdm/read_write/netcdf/flatten/flatten.py
- Rename "file" to "dataset"
- Use `match` clauses to switch between dataset format APIs
- Changes to allow Zarr datasets to be flattened
  - Zarr datasets to not have well-defined Dimensions, so finding
    which dimensions belong to which groups for Zarr datasets is more
    involved, and needs configuring with the new
    `group_dimension_search` keyword.
#### cfdm/read_write/netcdf/netcdfread.py
- Rename "file" to "dataset"
- Use `match` clauses to switch between dataset format APIs
- Handle Zarr shards
- Performance improvements in `_cache_data_elements`, largely aimed at
  reducing the number of requests to disk
#### cfdm/read_write/netcdf/netcdfwrite.py
- Rename "file" to "dataset"
- Use `match` clauses to switch between dataset format APIs
- Handle Zarr shards
#### cfdm/read_write/netcdf/zarr.py
- Include a reference variable in `ZarrDimension`
#### cfdm/read_write/read.py
- New keywords `store_dataset_shards` and `group_dimension_search`
#### cfdm/read_write/write.py
- Rename "file" to "dataset"
- New keyword `dataset_shards`
#### setup.py
- `zarr` import optional

### DIFF
--- a/cfdm/read_write/netcdf/netcdfwrite.py
+++ b/cfdm/read_write/netcdf/netcdfwrite.py
@@ -5028,11 +5028,6 @@ class NetCDFWrite(IOWrite):
                     raise RuntimeError(f"{error}: {dataset_name}")
 
             case "zarr":
-                if mode == "a":
-                    raise ValueError(
-                        "Can't write with mode 'a' to a Zarr dataset"
-                    )
-
                 try:
                     import zarr
                 except ModuleNotFoundError as error:
@@ -5560,6 +5555,11 @@ class NetCDFWrite(IOWrite):
         effective_fields = fields
 
         if mode == "a":
+            if fmt == "ZARR3":
+                raise ValueError(
+                    "Can't write with mode 'a' to a Zarr dataset"
+                )
+            
             # First read in the fields from the existing dataset:
             effective_fields = self._NetCDFRead(self.implementation).read(
                 dataset_name, netcdf_backend="netCDF4"
--- a/cfdm/test/test_zarr.py
+++ b/cfdm/test/test_zarr.py
@@ -136,7 +136,7 @@ class read_writeTest(unittest.TestCase):
             self.assertEqual(z["q"].shards, (4, 6))
 
     def test_zarr_read_write_CFA(self):
-        """Test CF aggreagtion in Zarr."""
+        """Test CF aggregation in Zarr."""
         f = self.f0
 
         cfdm.write(f, tmpdir1, fmt="ZARR3")
@@ -157,6 +157,14 @@ class read_writeTest(unittest.TestCase):
         self.assertTrue(z.equals(f))
         self.assertTrue(z.equals(n))
 
+    def test_zarr_write_append(self):
+        """Test in append mode with Zarr."""
+        # Check that append mode does not work for Zarr
+        f = self.f0
+        cfdm.write(f, tmpdir1, fmt='ZARR3')
+        with self.assertRaises(ValueError):
+            cfdm.write(f, tmpdir1, fmt='ZARR3', mode="a")
+            
     def test_zarr_groups_1(self):
         """Test for the general handling of Zarr hierarchical groups."""
         f = cfdm.example_field(1)


cfdm/data/netcdfindexer.py
```diff
@@ -623,7 +629,10 @@ def _mask(self, data, dtype, attributes, dtype_unsigned_int):
             if fvalisnan:
                 mask = np.isnan(data)
             else:
-                mask = data == fval
+                # Must use `np.asanyarray` here, to ensure that that
+                # 'mask' is a never a `bool`, which would make the
+                # following 'mask.any' call' fail.
```
sadielbartholomew:
```suggestion
                # following 'mask.any' call fail.
```

docs/source/installation.rst
```diff
@@ -237,6 +240,19 @@ The cfdm package requires:
   then setting the ``UDUNITS2_XML_PATH`` environment variable is the
   likely solution.
 
+Optional
+^^^^^^^^
+
+Some further dependencies that enable further functionality are
+optional. This to facilitate cfdm being installed in restricted
```
sadielbartholomew:
```suggestion
optional. This is to facilitate cfdm being installed in restricted
```

cfdm/read_write/netcdf/zarr.py
```diff
@@ -37,6 +43,34 @@ def __len__(self):
         """
         return self.size
 
+    def __repr__(self):
+        """The size of the dimension.
+
+        x.__len__() <==> len(x)
```
sadielbartholomew:
```suggestion
        x.__repr__() <==> repr(x)
```

cfdm/read_write/write.py
```diff
@@ -6,21 +6,23 @@
...
-    **NetCDF variable and dimension names**
+    **Dataset variable and dimension names**
 
-    These names are stored within constructs read a from dataset, or
-    may be set manually. They are used when writing a field construct
-    to the file. If a name has not been set then one will be
-    constructed (usually based on the standard name if it exists). The
-    names may be modified internally to prevent duplication in the
-    file.
+    These names are stored within constructs and are either read a
```
```suggestion
    These names are stored within constructs and are either read
```

cfdm/read_write/write.py
```diff
@@ -146,19 +148,20 @@ class write(ReadWrite):
...
             ========  =================================================
-            ``'w'``   Open a new file for writing to. If it exists and
-                      *overwrite* is True then the file is deleted
-                      prior to being recreated.
+            ``'w'``   Open a new dataset for writing to. If it exists
+                      and *overwrite* is True then the dataset is
+                      deleted prior to being recreated.
 
-            ``'a'``   Open an existing file for appending new
+            ``'a'``   Open an existing dataset for appending new
```
sadielbartholomew: Have you tried 'append' mode with Zarr?

cfdm/read_write/write.py
```diff
@@ -611,6 +617,44 @@ class write(ReadWrite):
...
+
+            * `None`
+
+               No sharding.
+
+            * `int`
+
+              The integer number of dataset chunks to be stored in a
+              single shard, favouring an equal number of dataset
+              chunks along each shard dimenson.
```
```suggestion
              chunks along each shard dimension.
```

cfdm/functions.py
```diff
@@ -2430,6 +2430,18 @@ def parse_indices(shape, indices, keepdims=True, newaxis=False):
...
 
 
+def netcdf_flatten(*args, **kwargs):
+    """Create a flattened version of a grouped CF dataset.
+
+    .. versionadded:: (cfdm) 1.11.2.0
+
+    """
+    raise DeprecationError(
+        "Function 'netcdf_flatten' has been renamed 'dataset_flatten' "
```
sadielbartholomew: This is probably worth mentioning in the Changelog as it concerns an API change, It is of course related to the Issues mentioned there already, but is a distinct thing so good to clarify separately, I think.

cfdm/data/aggregatedarray.py
```diff
@@ -262,9 +262,17 @@ def _parse_fragment_array(self, aggregated_filename, fragment_array):
                 if not scalar:
                     identifier = fa_identifiers[index].item()
 
+                uri = fa_uris[index]
+                try:
+                    # 'uri' is scalar numpy string type
+                    uri = uri.item()
+                except AttributeError:
+                    # E.g. 'uri' is a `str` instance
```
sadielbartholomew:
```suggestion
                    # E.g. 'uri' is already a `str` instance
```

or a similar comment somewhere nearby to make it clear that's the form we want, maybe?

cfdm/read_write/netcdf/netcdfread.py
```diff
@@ -8181,44 +8213,57 @@ def _create_Data(
...
+                    string_type = isinstance(array, str)
+                    if string_type:
+                        # A netCDF string type scalar variable comes
+                        # out as Python str object, so convert it to a
+                        # numpy array.
+                        array = np.array(array, dtype=f"U{len(array)}")
+
+                    if not variable.ndim:
+                        # NetCDF4 has a thing for making scalar size 1
+                        # variables into 1d arrays
```
sadielbartholomew: 😆

cfdm/read_write/netcdf/flatten/flatten.py
```diff
@@ -903,55 +1121,81 @@ def increment_pos(self, pos, dim, copy_slice_shape, var_shape):
             # increment. Finish.
             return False
 
-    def write_data_in_chunks(self, old_var, new_var):
-        """Copy the data of a variable to a new one by slice.
+    def write_data(self, old_var, new_var):
+        """Copy the data of a variable to the ouput datset.
```
sadielbartholomew:
```suggestion
        """Copy the data of a variable to the output dataset.
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -66,28 +73,40 @@ def cf_cell_method_qualifiers(self):
...
-        """Creates a new netCDF4 group object.
+    def _createGroup(self, parent, group_name):
+        """Creates a new dataset group object.
 
         .. versionadded:: (cfdm) 1.8.6.0
 
         :Parameters:
 
-            nc: `netCDF4._netCDF4.Group` or `netCDF4.Dataset`
+            parent: `netCDF4.Dateset` or `netCDF4.Group` or `Zarr.Group`
```
```suggestion
            parent: `netCDF4.Dataset` or `netCDF4.Group` or `Zarr.Group`
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -2345,10 +2427,10 @@ def _write_field_ancillary(
         return ncvar
 
     def _write_cell_measure(self, f, key, cell_measure):
-        """Write a cell measure construct to the netCDF file.
+        """Write a cell measure construct to the dataset.
 
-        If an identical construct has already in the file then the cell
-        measure will not be written.
+        If an identical construct has already in the dataset then the
```
```suggestion
        If an identical construct is already in the dataset then the
```

(Again not something written in the PR, but might as well correct now.)

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -2484,26 +2572,133 @@ def _create_external(
...
+
+        match g["backend"]:
+            case "netCDF4":
+                netcdf4_kwargs = kwargs
+                if "dimensions" not in kwargs:
+                    netcdf4_kwargs["dimensions"] = ()
+
+                contiguous = kwargs.get("contiguous")
+
+                NETCDF4 = g["dataset"].data_model.startswith("NETCDF4")
```
sadielbartholomew: Ideally we can rename this as something lower case to prevent confusion about capitalised variables as constants in Python. Unless you are deliberately going against that convention for some good reason?

sadielbartholomew (approved):
Great stuff! With a satisfying amount of match-case additions 🙂 

Only minor comments, though note especially the more significant one about append mode, and as a general comment when I run `test_zarr` with Python 3.12 and Zarr 3.0.8 I get spammed with a warning of:

```
/home/slb93/miniconda3/envs/cf-env-312-numpy2/lib/python3.12/site-packages/zarr/codecs/vlen_utf8.py:44: UserWarning: The codec `vlen-utf8` is currently not part in the Zarr format 3 specification. It may not be supported by other zarr implementations and may change in the future.
  return cls(**configuration_parsed)
```

but when I run it with Python 3.13 and Zarr 3.1.0 I don't see those. So I suspect the Zarr version 3.0.8 might be to blame. Ideally we can check this isn't emerging as a result of our code and fix it if it is.

Otherwise all good so please merge once you've considered all of my feedback here.

cfdm/docstring/docstring.py
```diff
@@ -1313,6 +1383,20 @@
...
+    "{{sharding description}}": """
+        When writing to a Zarr dataset, sharding provides a mechanism
+        to store multiple dataset chunks in a single storage object or
+        file. Without sharding, each dataset chunk is written to its
+        own file. Traditional file systems and object storage systems
+        may have performance issues storing and accessing large number
+        of files, and small files can be inefficient to store if they
+        are smaller than the block size of the file system. Sharding
+        can improve performance by creating fewer, and larger, files
+        for storing the dataset chunks.
```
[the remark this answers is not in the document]
davidhassell: Largely culled from the `zarr` docs, so thanks to them :)

cfdm/data/zarrarray.py
```diff
@@ -98,18 +98,30 @@ def close(self, dataset):
...
-
-        return super().open(zarr.open, mode="r")
+        try:
+            import zarr
+        except ModuleNotFoundError as error:
+            error.msg += (
+                ". Install the 'zarr' package "
+                "(https://pypi.org/project/zarr) to read Zarr datasets"
+            )
+            raise
```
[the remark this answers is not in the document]
sadielbartholomew: Ah OK, I see what happened. Sorry, you are right so ignore my comment - it does work for `ModuleNotFoundError`. But interestingly, it does raise an `AttributeError` for other forms of exception errors, and therefore I assumed all for consistency (but had I thought to just attempt do an actual `import` like you have shown here that would have been much smarter and a directly comparable robust test!):


```python
>>> try:
...     ddd
... except NameError as error:
...     error.msg += " Etc.!"
...     raise
... 
  File "<stdin>", line 2, in <module>
NameError: name 'ddd' is not defined

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
AttributeError: 'NameError' object has no attribute 'msg'
```

Turns out the reason is [quoting from AI tool which didn't realise this until I pointed it out, naturally, and helped to lead me astray on the above]:

* "ModuleNotFoundError inherits from ImportError, and ImportError is one of the few built-in exceptions that explicitly defines a msg attribute."
* "Python has never standardized .msg across exceptions, largely for backward-compatibility reasons."

So, I learnt something here! Sorry again for the dud comment. Will mark this and the other comparable comment as resolved.

cfdm/read_write/netcdf/netcdfread.py
```diff
@@ -707,11 +714,14 @@ def _open_zarr(self, dataset):
         try:
             import zarr
         except ModuleNotFoundError as error:
-            error.msg += ". Install the 'zarr' package to read Zarr datasets"
+            error.msg += (
+                ". Install the 'zarr' package "
+                "(https://pypi.org/project/zarr) to read Zarr datasets"
+            )
```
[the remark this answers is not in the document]
sadielbartholomew: Indeed. Turns out I was wrong and it works as per my comment on the other comment. Marking as resolved.

cfdm/read_write/netcdf/netcdfread.py
```diff
@@ -445,45 +445,46 @@ def _reference(self, ncvar, referencing_ncvar):
...
-            # zarr
-            return
-
-        # netCDF4, h5netcdf
         for nc in g["datasets"]:
-            nc.close()
+            try:
+                nc.close()
+            except AttributeError:
+                pass
```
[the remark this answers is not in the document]
sadielbartholomew: Oh very good point! I should have deduced that from the specific Exception raised 🤦🏻‍♀️ . Marking as resolved.

cfdm/test/test_zarr.py
```diff
@@ -0,0 +1,333 @@
...
+        cfdm.LOG_LEVEL("DISABLE")
+        # Note: to enable all messages for given methods, lines or
+        # calls (those without a 'verbose' option to do the same)
+        # e.g. to debug them, wrap them (for methods, start-to-end
+        # internally) as follows: cfdm.LOG_LEVEL('DEBUG')
+        #
+        # < ... test code ... >
+        # cfdm.log_level('DISABLE')
+
+    def test_zarr_read_write_1(self):
```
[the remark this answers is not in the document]
davidhassell: Good call. How about: 769c2a996e7074f1ab2dc633a1d5c6440332e175

PR #371, opened 2025-11-27
Implement HEALPix grids

### LINKED ISSUE #370: Implement HEALPix grids
As of CF-1.13, HEALPix grids can be represented in CF.

`cfdm` needs modifying to recognise these grids.

davidhassell:
Fixes #370


sadielbartholomew (approved):
All good and consistent with the [CF Conventions documenting of the new HEALPix grid mapping](https://cfconventions.org/cf-conventions/cf-conventions.html#healpix). The two new example fields should be useful!

Only one minor note for us to think about for future. Please merge.

PR #372, opened 2025-11-27
Implementation of UGRID (part 2 of 2): Write

### LINKED ISSUE #271: Implementation of UGRID (part 2 of 2)
CF-1.11 will recognise the UGRID  conventions (https://github.com/cf-convention/cf-conventions/issues/153), and so cfdm needs enhancing to implement the extended CF data model that this implies (specifically, the new Domain Topology and Cell Connectivity constructs).

Part 1 (https://github.com/NCAS-CMS/cfdm/issues/270) of this enhancement deals with the internal representation and reading of datasets.

This part (2) will implement the ability to write UGRID datasets to disk.

davidhassell:
Fixes #271 

Requires #371 (which has now been merged into main)


cfdm/read_write/netcdf/netcdfwrite_ugrid.py
```diff
@@ -0,0 +1,1010 @@
...
+                domain_topology,
+                domain_axes,
+            )
+
+        g["key_to_ncvar"][key] = ncvar
+        g["key_to_ncdims"][key] = ncdimensions
+
+        return ncvar
+
+    def _ugrid_write_cell_connectivity(self, f, key, cell_connectivity):
```
sadielbartholomew: This method is pretty logically similar to `_ugrid_write_domain_topology` - for now it's fine to define them separately but in future especially if we add any similar methods we can consolidate them to all use a`write_to_connectivity_variable` method or something along those lines?

sadielbartholomew (changes requested):
Looks great, other than one bit of code dodginess in `netcdfread` that I've highlighted in-line which needs to be updated before I'm happy to merge. Otherwise just the standard minor comments, mostly some typos/formatting issues.

Please consider my feedback and then we should be able to quickly move towards merging. Thanks.

sadielbartholomew (approved):
Thanks for addressing the feedback, especially on the main issue https://github.com/NCAS-CMS/cfdm/pull/372#discussion_r3015999268. I have done a final sanity check and everything looks good so please merge!

PR #376, opened 2026-01-06
Control setting and display of cached data elements 

### LINKED ISSUE #363: Control setting and display of cached data elements
When the array being passed to `cfdm.Data.__init__` is not a Dask array and is in memory, why don't we grab the first (, second) and last values for the cache, as it's much quicker to do it at this time than hauling them out of the instantiated Dask array at a later stage.

PR to follow.

davidhassell:
Fixes #363

### DIFF
--- a/cfdm/read_write/netcdf/netcdfread.py
+++ b/cfdm/read_write/netcdf/netcdfread.py
@@ -1979,7 +1979,7 @@ class NetCDFRead(IORead):
                 "        read_vars['parsed_aggregated_data'] =\n"
                 f"            {g['parsed_aggregated_data']}"
             )  # prgama: no cover
-                
+
         # ------------------------------------------------------------
         # List variables
         #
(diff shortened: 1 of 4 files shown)


cfdm/data/data.py
```diff
@@ -992,10 +1017,35 @@ def __setitem__(self, indices, value):
...
+                getting these data elements needs an expensive
+                computation, possibly including a slow read from local
+                or remote disk.
+
+                If False then do not show such data elements, *unless
+                data elements have been previously cached*, thereby
+                avoiding a potentially high computational cost.
+
+                If `None` (the default) then the value of *data* will
+                taken from the `{{package}}.display_data` function.
```
sadielbartholomew:
```suggestion
                be taken from the `{{package}}.display_data` function.
```

cfdm/data/data.py
```diff
@@ -5345,6 +5591,32 @@ def get_deterministic_name(self):
             units._canonical_calendar,
         )
 
+    def get_cached_elements(self):
+        """Get the cache of selected array elements.
+
+        If the cache is empty, then `cache_elements` may used to
+        populated it.
```
sadielbartholomew:
```suggestion
        populate it.
```

PR #387, opened 2026-02-24
Fix aggregation unlimited netCDF dimensions

davidhassell:
Fix bug in `cfdm.write` to correctly create unlimited aggregation dimensions that correspond to unlimited normal dimensions


jonnyhtw (commented):
Confirmed working using data from the [CANARI](https://canari.ac.uk/) climate model ensemble. 

Worth noting just for reference that without this fix, variables can be promoted to `UNLIMITED` using the `--mk_rec_dmn` option to [`ncks`](https://linux.die.net/man/1/ncks).

PR #384, opened 2026-02-18
Enable parallel netCDF dataset reads

davidhassell:
Fixes #375

PR #388, opened 2026-02-25
New 'h5netcdf-h5py' netCDF backend in `cfdm.write`, and 'h5py_options' parameter 

### LINKED ISSUE #386: New 'h5netcdf-h5py' netCDF backend in `cfdm.write`, and 'h5py_options' parameter
By introducing the use of `h5netcdf` with the `h5py` backend as a backend option to `cfdm.write`, we open up being able to control the HDF5 internal metadata structure via the full range of `h5py.File` keyword arguments (https://docs.h5py.org/en/stable/high/file.html#h5py.File), which can be set with the new `h5py_option` parameter to `cfdm.write`.

For now, the `h5py_option` parameter will be the only way to control these option, but in the future we could make some common options available directly through the `cfdm` API. 

Also in the future, we may well want to set some of these HD5 options automatically (be default) based on the nature of the fields being written out.

PR to follow.

davidhassell:
Fixes #386

Is built on top of #384, which must be concluded before this PR is considered. Changes in #384 arising from its review must be merged in this PR before its own review.


cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -2612,6 +2653,9 @@ def _create_external(
     def _createVariable(self, **kwargs):
         """Create a variable in the dataset.
 
+        Each backend needs a seperate parsing the input kwargs to suit
```
```suggestion
        Each backend needs a separate parsing of the input kwargs to suit
```

(Typo fix plus there seems to be a missing word. I have guessed it.)

cfdm/test/test_read_write.py
```diff
@@ -1384,6 +1409,64 @@ def test_write_differing_formula_terms(self):
...
+        # Bad fmt/backend combinations
+        for backend in ("netCDF4", "h5netcdf-h5py"):
+            with self.assertRaises(ValueError):
+                cfdm.write(f, tmpfile, fmt="ZARR3", netcdf_backend=backend)
+
+        for backend in ("zarr", "h5netcdf-h5py"):
+            with self.assertRaises(ValueError):
+                cfdm.write(
+                    f, tmpfile, fmt="NETCDF3_CLASSIC", netcdf_backend=backend
+                )
```
sadielbartholomew: Ideally we can add a few assertions that we error elegantly when an append mode is attempted with any backend other than `"netCDF4"`:

```suggestion
            with self.assertRaises(ValueError):
                cfdm.write(
                    f, tmpfile, fmt="NETCDF3_CLASSIC", netcdf_backend=backend
                )
            # Check also that append mode attempt is caught
            # (is only valid for netCDF4 backend)
            with self.assertRaises(ValueError):
                cfdm.write(f, tmpfile, netcdf_backend=backend)
```

(This could live in the `test_write_netcdf_mode` just as well, if you prefer, but it was easier to 'suggest' it here given the code additions already set up and it would also work here.)

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -5548,6 +5733,60 @@ def write(
...
+        # Set dataset_fmt
+        match fmt:
+            case "ZARR3":
+                self.write_vars["dataset_type"] = "directory"
+            case _:
+                self.write_vars["dataset_type"] = "file"
+
+        # Parse the 'h5py_options' parameter
+        if h5py_options is None:
+            self.write_vars["h5py_options"] = {}
```
sadielbartholomew: (Perhaps as an `else` clause to this, but could go anywhere ideally nearer the start of the function to catch early) it might be nice to at least warn via the logging system, but arguably error instead, if some `h5py_options` are specified but the backend isn't `h5netcdf-h5py`? (Apologies if I've missed some logic that does that somewhere.)

sadielbartholomew (approved):
Does what it says on the tin! I have minor comments as in-line but otherwise all good. Please merge when ready.

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -5548,6 +5733,60 @@ def write(
...
+            raise ValueError(
+                "Can only set omit_data=True when netcdf_backend='netCDF4'"
+            )
+
+        # Set dataset_fmt
+        match fmt:
+            case "ZARR3":
+                self.write_vars["dataset_type"] = "directory"
+            case _:
+                self.write_vars["dataset_type"] = "file"
```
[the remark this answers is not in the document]
davidhassell: Not confident enough to keep the match statement :)

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -5548,6 +5733,60 @@ def write(
...
+                case fmt if fmt in NETCDF4_FMTS:
+                    backend = "h5netcdf-h5py"
+                case fmt if fmt in ZARR_FMTS:
+                    backend = "zarr"
+                case fmt if fmt in NETCDF3_FMTS:
+                    backend = "netCDF4"
+
+            self.write_vars["backend"] = backend
+
+        match backend:
```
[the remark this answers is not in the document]
davidhassell: In general I don't like pre-defining f-strings, as they can sometimes contain large overheads (particularly in read/write).... not in this case :), but nonetheless ....

PR #399, opened 2026-03-20
Convert to `xarray`

### LINKED ISSUE #394: Convert to `xarray`
It would be nice to be able to convert `cfdm.Field` and `cfdm.Domain` constructs to `xarray` datasets in memory.

E.g.
```python
>>> import cfdm
>>> f = cfdm.example_field(0)
>>> ds = f.to_xarray()
>>> ds
<xarray.Dataset> Size: 640B
Dimensions:   (lat: 5, lon: 8, bounds2: 2)
Coordinates:
  * lat       (lat) float64 40B -75.0 -45.0 0.0 45.0 75.0
  * lon       (lon) float64 64B 22.5 67.5 112.5 157.5 202.5 247.5 292.5 337.5
    time      float64 8B dask.array<chunksize=(), meta=np.ndarray>
Dimensions without coordinates: bounds2
Data variables:
    lat_bnds  (lat, bounds2) float64 80B dask.array<chunksize=(5, 2), meta=np.ndarray>
    lon_bnds  (lon, bounds2) float64 128B dask.array<chunksize=(8, 2), meta=np.ndarray>
    q         (lat, lon) float64 320B dask.array<chunksize=(5, 8), meta=np.ndarray>
Attributes:
    Conventions:  CF-1.13
>>>
```
```python
>>> f.nc_set_variable('/forecast/model/q')
>>> ds = f.to_xarray()
>>> ds 
<xarray.DataTree>
Group: /
│   Dimensions:   (lat: 5, lon: 8, bounds2: 2)
│   Coordinates:
│     * lat       (lat) float64 40B -75.0 -45.0 0.0 45.0 75.0
│     * lon       (lon) float64 64B 22.5 67.5 112.5 157.5 202.5 247.5 292.5 337.5
│       time      float64 8B dask.array<chunksize=(), meta=np.ndarray>
│   Dimensions without coordinates: bounds2
│   Data variables:
│       lat_bnds  (lat, bounds2) float64 80B dask.array<chunksize=(5, 2), meta=np.ndarray>
│       lon_bnds  (lon, bounds2) float64 128B dask.array<chunksize=(8, 2), meta=np.ndarray>
│   Attributes:
│       Conventions:  CF-1.13
└── Group: /forecast
    └── Group: /forecast/model
            Dimensions:  (lat: 5, lon: 8)
            Data variables:
                q        (lat, lon) float64 320B dask.array<chunksize=(5, 8), meta=np.ndarray>
>>>
```
```python
>>> ds = cfdm.write(f, fmt='XARRAY')
>>> ds
<xarray.DataTree>
Group: /
│   Dimensions:   (lat: 5, lon: 8, bounds2: 2)
│   Coordinates:
│     * lat       (lat) float64 40B -75.0 -45.0 0.0 45.0 75.0
│     * lon       (lon) float64 64B 22.5 67.5 112.5 157.5 202.5 247.5 292.5 337.5
│       time      float64 8B dask.array<chunksize=(), meta=np.ndarray>
│   Dimensions without coordinates: bounds2
│   Data variables:
│       lat_bnds  (lat, bounds2) float64 80B dask.array<chunksize=(5, 2), meta=np.ndarray>
│       lon_bnds  (lon, bounds2) float64 128B dask.array<chunksize=(8, 2), meta=np.ndarray>
│   Attributes:
│       Conventions:  CF-1.13
└── Group: /forecast
    └── Group: /forecast/model
            Dimensions:  (lat: 5, lon: 8)
            Data variables:
                q        (lat, lon) float64 320B dask.array<chunksize=(5, 8), meta=np.ndarray>
>>>
```

davidhassell:
Fixes #394 

~Needs #372 to be merged, first~ (_edit 2026-06-09 - which it now is_)


cfdm/mixin/fielddomain.py
```diff
@@ -2624,3 +2624,38 @@ def set_mesh_id(self, mesh_id):
...
+        """Convert the {{class}} to an `xarray` dataset.
+
+        {{cf_xarray description}}
+
+        Note that ``ds = f.to_xarray()`` is identical to ``ds =
+        cfdm.write(f, fmt='XARRAY')``; and multiple {{class_lower}}s
+        may be written to the same `xarray` dataset with `cfdm.write`
+        (e.g. ``ds = cfdm.write([f, g], fmt='XARRAY')``). Also,
+        `cfdm.write` allows a mixture a mixture of fields and domains
+        to be written to the same `xarray` dataset.
```
sadielbartholomew, line 2638: Appending an 's' to a class name is a bit dangerous as not everything becomes plural that way, so to be safe I'd avoid having to form the plural via something like:
```suggestion
        Note that ``ds = f.to_xarray()`` is identical to ``ds =
        cfdm.write(f, fmt='XARRAY')``; and multiple instances of
        {{class_lower}} may be written to the same `xarray` dataset
        with `cfdm.write` (e.g. ``ds = cfdm.write([f, g], fmt='XARRAY')``).
        Also, `cfdm.write` allows a mixture a mixture of fields and
        domains to be written to the same `xarray` dataset.
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -3068,6 +3085,14 @@ def _write_netcdf_variable(
         if not construct_type:
             construct_type = self.implementation.get_construct_type(cfvar)
 
+        # Whether or not the variable is a coordinate variable (of any
+        # type). Bounds variables, interior rung variables, etc, are
```
```suggestion
        # type). Bounds variables, interior ring variables, etc., are
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -5596,9 +5657,9 @@ def write(
 
         logger.info(f"Writing to {fmt}")  # pragma: no cover
 
-        # Expand dataset name
-        dataset_name = os.path.expanduser(os.path.expandvars(dataset_name))
-        dataset_name = abspath(dataset_name)
+        #        # Expand dataset name
+        #        dataset_name = os.path.expanduser(os.path.expandvars(dataset_name))
+        #        dataset_name = abspath(dataset_name)
```
sadielbartholomew: To tidy / avoid clutter:

```suggestion
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -5772,6 +5838,52 @@ def write(
...
+        if fmt in ZARR_FMTS:
+            self.write_vars["dataset_type"] = "directory"
+        elif fmt in XARRAY_FMTS:
+            # Writing to memory, not to disk.
+            self.write_vars["dataset_type"] = "memory"
+            self.write_vars["write_to_disk"] = False
+        else:
+            self.write_vars["dataset_type"] = "file"
+
+        # Parse the data set name
```
```suggestion
        # Parse the dataset name
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -5772,6 +5838,52 @@ def write(
...
+                )
+
+            # Expand dataset name
+            dataset_name = os.path.expanduser(os.path.expandvars(dataset_name))
+            dataset_name = abspath(dataset_name)
+            self.write_vars["dataset_name"] = dataset_name
+        elif dataset_name is not None:
+            # Must not provide a dataset name when not writing to disk
+            raise ValueError(
+                "Must not provide a local dataset name when writing to "
```
sadielbartholomew: Not essential but the word ordering seems a bit clunky here (fine for the comment but ideally not user-facing messages) so I'd change the order to:

```suggestion
                "A local dataset name must not be provided when writing to "
```

cfdm/read_write/netcdf/xarray_dataset.py
```diff
@@ -0,0 +1,300 @@
+class XarrayDataset:
+    """An `xarray` dataset` constructor.
```
sadielbartholomew:
```suggestion
    """An `xarray` dataset constructor.
```

cfdm/read_write/write.py
```diff
@@ -147,6 +158,15 @@ class write(ReadWrite):
             ``'NETCDF4'`` files use the version 4 disk format (HDF5)
             and use the new features of the version 4 API.
 
+            ``'XARRAY'`` datasets in memory will be either
+            `xarray.Dataset` (if there are no sub-groups of the root
+            group) or else `xarray.DataTree` (if there are sub-groups
+            of the root group).If the `cf_xarray` package
```
```suggestion
            of the root group). If the `cf_xarray` package
```

cfdm/test/test_xarray.py
```diff
@@ -0,0 +1,137 @@
...
+            "geometry_4.nc",
+            "geometry_interior_ring.nc",
+            "geometry_interior_ring_2.nc",
+            "string_char.nc",
+            "subsampled_2.nc",
+            "ugrid_1.nc",
+            "ugrid_2.nc",
+            "ugrid_3.nc",
+            "test_file.nc",
+        ):
```
sadielbartholomew, line 85: Very nice - testing over a large sample of our netCDF files of various forms and all the example fields as you've done should endure good coverage. +1

sadielbartholomew (commented):
Does what it says on the tin and in a direct sensible way. That said, we should definitely work towards more comprehensive testing than just asking 'is it a valid xarray dataset/tree?' after conversion to confirm that the conversion maps everything correctly, but given how many datasets including example fields are being tested on, that would take a while. Let's open a follow-on Issue to capture that we should test on that rigorously?

Given this, to review I printed out the various fields and their converted equivalent and had a look to see how they compared. I was wondering about the following, which may be by design but may be issues:
1. Some size 1 dimension coordinates are registered as scalar coordinates only in the resulting `Dataset` - see e.g. the 0th example field (which has `Dimension coordinate: time`) conversion:

   ```python
     <xarray.Dataset> Size: 640B
    Dimensions:   (lat: 5, lon: 8, bounds2: 2)
    Coordinates:
      * lat       (lat) float64 40B -75.0 -45.0 0.0 45.0 75.0
      * lon       (lon) float64 64B 22.5 67.5 112.5 157.5 202.5 247.5 292.5 337.5
        time      float64 8B dask.array<chunksize=(), meta=np.ndarray>
    Dimensions without coordinates: bounds2
    Data variables:
        lat_bnds  (lat, bounds2) float64 80B dask.array<chunksize=(5, 2), meta=np.ndarray>
        lon_bnds  (lon, bounds2) float64 128B dask.array<chunksize=(8, 2), meta=np.ndarray>
        q         (lat, lon) float64 320B dask.array<chunksize=(5, 8), meta=np.ndarray>
    Attributes:
        Conventions:  CF-1.13
    ```

2. Global properties, except for the `Conventions` value, appear to be missing on the converted xarray, with just:
    ```    ...
    Attributes:
        Conventions:  CF-1.13 
    ```
    showing (see all print outs below). We should just need to attach these to the `ds.attrs` dict, right?

3. There may be an issue with time decoding. Even with `cf_xarray` installed, I can't get datetimes to be interpreted, for example `ds.time.values` gives float values.

Would be good to get your thoughts on these before I approve. Otherwise only minor things e.g. typos (see inline comments).

#### Helper

For ease of comparison of the results, I used the below - can apply from the `diff` for ease:

```diff
diff --git a/cfdm/test/test_xarray.py b/cfdm/test/test_xarray.py
index 4d59ec2f3..d15e76fff 100644
--- a/cfdm/test/test_xarray.py
+++ b/cfdm/test/test_xarray.py
@@ -32,6 +32,8 @@ class xarrayTest(unittest.TestCase):
         # Write each field to a different xarray dataset
         for f in fields:
             ds = f.to_xarray()
+            print(f"HAVE EX FIELD:\n", ds, "\nfor")
+            f.dump()
             self.assertIsInstance(ds, xr.Dataset)
             self.assertIn("Conventions", ds.attrs)
             str(ds)
@@ -85,6 +87,8 @@ class xarrayTest(unittest.TestCase):
         ):
             for f in cfdm.read(dataset):
                 ds = f.to_xarray()
+                print("HAVE NC FIELD:\n", ds, "\nfor")
+                f.dump()
                 self.assertIsInstance(ds, xr.Dataset)
                 str(ds)
```

cfdm/mixin/fielddomain.py
davidhassell: I think this is OK - `{{class_lower}}` is not acting as  class name (which would be `{{class}}`- but a plain text construct descriptor: _"and multiple fields may be written to the same `xarray` dataset"_

davidhassell:
> 1. Some size 1 dimension coordinates are registered as scalar coordinates only in the resulting Dataset - see e.g. the 0th example field (which has Dimension coordinate: time) conversion:

Scalar coordinates are size one. In cfdm (and the CF data model)  the concept of "scalar coordinate" does not exist. We know that to write them as scalar (i.e. zero dimensional) _in an encoding_ (xarray in this case) because their domain axis is not spanned by the fields data array, which is the case for the time coordinate here.

> 2. Global properties, except for the Conventions value, appear to be missing on the converted xarray, with just:

I don't think they're missing - that field only has one global attribute:
```python
>>> f.nc_global_attributes()
{'Conventions': None}
```

> 3. There may be an issue with time decoding. Even with cf_xarray installed, I can't get datetimes to be interpreted, for example ds.time.values gives float values.

I'd not seen that! Will fix in an upcoming commit: 7d9c5a1c01e3aa8a7e29ce9186c87810d71624f7

davidhassell:
> Let's open a follow-on Issue to capture that we should test on that rigorously?

Good idea.

sadielbartholomew:
> that field only has one global attribute

Ah sorry, I meant to specify this is seen from example field 1, which has `project = 'research'` but that doesn't come through in the `g.attrs` where `g = cfdm.example_field(1).to_xarray()`. Though I also don't see that in `f.nc_global_attributes()` (which is an empty dict) for that example field so maybe I am misunderstanding what `project` key-value is CF wise?

davidhassell:
I think it's OK:
```python
In [2]: import cfdm; f = cfdm.example_field(0)

In [3]: x = f.to_xarray()

In [4]: x.attrs
Out[4]: {'Conventions': 'CF-1.13'}

In [5]: x.q.attrs
Out[5]: 
{'project': 'research',
 'standard_name': 'specific_humidity',
 'units': '1',
 'coordinates': 'time',
 'cell_methods': 'area: mean'}

```

sadielbartholomew:
> Scalar coordinates are size one. In cfdm (and the CF data model) the concept of "scalar coordinate" does not exist. We know that to write them as scalar (i.e. zero dimensional) in an encoding (xarray in this case) because their domain axis is not spanned by the fields data array, which is the case for the time coordinate here.

Ah yes, fair. I knew scalar coords are size 1 so didn't word my review comment very well, sorry. All good, as discussed offline.

> I think it's OK

Thanks for clarifying - the properties show at the top of the field `dump()` representation are not all global attributes as I thought they were, the source of the confusion. OK indeed!

setup.py
```diff
@@ -64,17 +64,18 @@ def _get_version():
...
-* modify field and domain construct metadata and data,
-* create subspaces of field and domain constructs,
-* incorporate, and create, metadata stored in external files, and
 * read, write, and create data that have been compressed by convention
   (i.e. ragged or gathered arrays, or coordinate arrays compressed
   by subsampling), whilst presenting a view of the data in its
   uncompressed form,
 * read and write that data that are quantized to eliminate false
   precision.
+* Convert field and domain constructs to `xarray` datasets in memory,
```
sadielbartholomew: For consistency:

```suggestion
* convert field and domain constructs to `xarray` datasets in memory,
```

sadielbartholomew (approved):
Thanks for incorporating the time decoding and for clarifying offline about the other review items. All feedback is now addressed, thanks (though I have one final minor wording thing from a sanity check, see comment). After that's in, please merge!

davidhassell:
Thanks, Sadie.

PR #400, opened 2026-03-22
Fix bug in `cfdm.netcdf_indexer` that sometimes caused a failure with a `np.newaxis` index

### LINKED ISSUE #395: Bug: Python 3.14: coordinate array materialization can fail with h5netcdf/h5py (TypeError: Indexing with None (or np.newaxis) is not supported)
am seeing a regression when materializing some coordinate arrays after reading/collapsing NetCDF data with cf-python/cfdm on Python 3.14.

Environment:

Python 3.14.3
cf-python 3.19.0
cfdm 1.13.0.0
h5netcdf 1.8.1
h5py 3.16.0
dask 2026.3.0
numpy 2.4.3
Observed behavior:

Field array materialization may succeed.
One coordinate array (in my case the height coordinate) fails with:
TypeError: Indexing with None (or np.newaxis) is not supported
Minimal repro:
```python
import cf

f = cf.read("test1.nc")[0]
p = f.collapse("time: mean", weights=True)

# Field materialization can succeed:
_ = p.array

# Coordinate materialization can fail for one coordinate:
for key, c in p.dimension_coordinates(todict=True).items():
    print("trying", key, c.identity(default="unknown"))
    arr = c.array  # fails for one coordinate in this environment
    print(arr.shape)
```
Full error line:
TypeError: Indexing with None (or np.newaxis) is not supported

Expected behaviour:

Coordinate array access should either work consistently for all dimension coordinates, or fail with a cfdm-level fallback/normalization that avoids passing unsupported index forms into h5py/h5netcdf.
Additional notes:

This appeared after moving to Python 3.14. In xconv2, this now occurs during coordinate metadata extraction and also on some collapsed plotting paths unless guarded.

`test1.nc` manifests this!

davidhassell:
Fixes #395


cfdm/test/test_functions.py
```diff
@@ -780,6 +780,14 @@ def test_dirname(self):
             "/data",
         )
 
+    def test_axis_dropping_axis(self):
```
sadielbartholomew: I think you meant to call it (or, would be the direct mapping choice):

```suggestion
    def test_axis_dropping_index(self):
```

sadielbartholomew (approved):
Makes the fix and adds suitable testing. I've also recreated the original issue using the MRE in https://github.com/NCAS-CMS/cfdm/issues/395#issuecomment-4119344068 and confirmed this PR fixes it.

Some typos (see in-line) but other than that all good, so please merge once those are in.

PR #411, opened 2026-07-10
Add dotfile cache support for standard names retrieval

sadielbartholomew:
...and generally improve and safeguard from network issues the handling of the fetching and storage of the names, e.g. store them as a frozenset rather than a list, which is more performant for membership testing and can't be edited by the user (whether by mistake, or otherwise!).

Specifically, this PR adds as a follow-on from #373:

* safe bypassing of standard names checks when the online XML can't be accessed (more likely but rare) or is not valid (rarer but possible);
* caching to a (pair, possibly of) local dotfiles under `~/.cf/` which store the names as a pickled frozen set and load them when required, including some basic cache metadata that will allow us to detect if a dotfile cache is old;
* timeouts on URL open attempts to prevent too much waiting on a resource (could be made configurable but for now I've set it as 5 seconds which seems sensible and proportionate given the size of the names XML).

And finally bundled here, since it is related to similar logic, is to move the checks to validate standard names after the checks that variables exists in all cases where the latter checks exist (that's the only sensible order there), all in [dfe1b0f](https://github.com/NCAS-CMS/cfdm/pull/411/commits/dfe1b0ff21970594db679b591a7f2fec3a047b90).

### DIFF
--- a/cfdm/conformance/__init__.py
+++ b/cfdm/conformance/__init__.py
@@ -5,11 +5,11 @@ from .datamodel import (
     NonConformanceCase,
     VariableNonConformance,
 )
+
+# Allow users to access TODO though not yet advertised (i.e. in the API ref)
+# - wait until conformance work is more mature before doing this
 from .standardnames import (
     get_all_current_standard_names,
-    # Intended for internal use but used in testing, so need exposing:
-    _extract_names_from_xml,
-    _STD_NAME_CURRENT_XML_URL,
 )
 
 # Eventually structure as with individual modules per class, something like:
--- a/cfdm/conformance/standardnames.py
+++ b/cfdm/conformance/standardnames.py
@@ -2,6 +2,7 @@ import datetime
 import logging
 import os
 import pickle
+import sys
 import xml.etree.ElementTree as ET
 from functools import lru_cache
 
@@ -29,8 +30,10 @@ DEFAULT_TIMEOUT_SECONDS = 5
 
 # Cache config.
 CACHE_DIR = ".cf"
-CACHE_PICKLE_FILENAME_NOALIASES = "standard_names.pickle"
-CACHE_PICKLE_FILENAME_WITHALIASES = "standard_names_with_aliases.pickle"
+# Filenames exclude the ".pickle" extension which is added later
+CACHE_PICKLE_FILENAME_NOALIASES = "standard_names"
+CACHE_PICKLE_FILENAME_WITHALIASES = "standard_names_with_aliases"
+
 CACHE_FORMAT_VERSION = 1
 CACHE_MAX_AGE_DAYS = 30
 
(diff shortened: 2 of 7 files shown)


cfdm/conformance/__init__.py
```diff
@@ -5,11 +5,11 @@
     NonConformanceCase,
     VariableNonConformance,
 )
+
+# Allow users to access TODO though not yet advertised (i.e. in the API ref)
+# - wait until conformance work is more mature before doing this
```
sadielbartholomew, line 10: As noted in this comment, I think that `get_all_current_standard_names` would be useful to users, allowing them to quickly query valid names, so I suggest that we add it to the user-facing API. But I haven't done so in this PR since it depends what you think David.

sadielbartholomew:
Sorry if you'd already started looking at this David - I just pushed up some (four) new commits for some wash-up, though that shouldn't confuse any reviewing since all they do is to (see commit messages for what is where):
* update some docstrings that I have since noticed weren't fully written (or didn't include parameters and return signature etc.);
* create the changelog entries for this PR;
* update one keyword arg. (`table_version`) to make it compulsory, which I noticed during doing the above.

cfdm/conformance/standardnames.py
```diff
@@ -46,28 +73,192 @@ def _extract_names_from_xml(snames_xml, include_aliases):
...
+    )
+
+    if not os.path.isdir(cache_dir):
+        os.makedirs(cache_dir)
+
+    if include_aliases:
+        filename = CACHE_PICKLE_FILENAME_WITHALIASES
+    else:
+        filename = CACHE_PICKLE_FILENAME_NOALIASES
+
```
davidhassell, line 155: Can we include the Python version in the file name, using similar notation to __pycache__, perhaps? (e.g. `standard_names-312.pickle`) - As far as I recall, pickle is not guaranteed to work across versions, and the cache can be up to 30 days old, so ....

davidhassell (approved):
OK - looks good. I confirm that all tests pass. One suggestion on the name of the pickle files.

cfdm/conformance/standardnames.py
sadielbartholomew: Ah yes - good thinking! My one concern was then that a user could end up with several cache files if users have different environments, so potentially say eight with `standard_names[_with_aliases]-[3X].pickle` for four versions of Python they might use at once (say, for testing), or much more over time given the gradual move to newer versions and lack of clearing of the cache. But I checked on the size of the files and see:

```console
$ ls -lrth ~/.cf/
total 604K
-rw-r--r-- 1 slb93 slb93 283K Jul 13 12:23 standard_names.pickle
-rw-r--r-- 1 slb93 slb93 318K Jul 13 12:23 standard_names_with_aliases.pickle
```

which is pretty small so even in the case inclusive of aliases at  ~300 KiB, per cache file that's not enough to become any issue even with hundreds.

I thought we could add the Python version into the cache metadata dictionary as well (or alternatively), but realised that is a bit pointless because you have to then read the cache with an appropriate Python version to access that information anyway.

So, that's all good - I will add a commit to update the naming as such.

sadielbartholomew: Done in [109a6ac](https://github.com/NCAS-CMS/cfdm/pull/411/commits/109a6ac955213017d43196452c37d18fe48ebcb1) (where I also tweaked a word here and there in the docs and changelog).

sadielbartholomew:
@davidhassell thanks for the review - feedback now addressed. Can I merge?

davidhassell:
Great - please merge :)

PR #415, opened 2026-07-27
Set sensible dataset chunksizes by default for 1-d data

### LINKED ISSUE #414: Set sensible dataset chunksizes by default for 1-d data
When reading a netCDF4 or Zarr  dataset, the performance can be seriously degrded if a data array has too many HDF5 chunks. This is particularly seen when 1-d coordinates (and their bounds) are chunked with one array element per chunk (e.g. commonly seen in CMIP6 datasets).

Whilst the dataset chunking can be controlled with the `nc_dataset_chunksize` method, and the `cfdm.write` `dataset_chunks` keyword, the current default behaviour is to retain the chunking read from disk when writing the field out to a new file on disk. This is fine, but it preserves any terrible chunking on the coordinates, which was _certainly_ not intended by the date writer.

In `cfdm.write`, we could easily override the `nc_dataset_chunksize` method, and the `cfdm.write` `dataset_chunks` keyword to ensure the sensible chunking of all 1-d arrays, and the bounds arrays of 1-d coordinates.

For such data, this would ensure that there is either one dataset chunk, or if the  default chunksize is greater than the limit specified by a new `cfdm.write` keyword argument (`one_d_chunks`), then that default chunksize can be used. This also allows the user to override the overriding!

By default the `one_d_chunks` should be 4 MiB, meaning that using `cfdm.write` will create more performant output files by default.

davidhassell:
Fixes #414

The idea is to add a bit of functionality to `NetCDFWrite._chunking_parameters` to modify the chunks it returns for 1-d arrays, and the bounds arrays of 1-d coordinates. There's also a new `cfdm.write` keyword `one_d_chunks` to control the behaviour.

### DIFF
--- a/cfdm/read_write/netcdf/netcdfwrite.py
+++ b/cfdm/read_write/netcdf/netcdfwrite.py
@@ -3045,7 +3045,7 @@ class NetCDFWrite(NetCDFWriteUgrid, IOWrite):
                 .. versionadded:: (cfdm) 1.12.0.0
 
             bounds: `bool`
-                Whether or not the data represent bounds.
+                Whether or not the data represent cell bounds.
 
                 .. versionadded:: (cfdm) NEXTVERSION
 
@@ -6498,7 +6498,9 @@ class NetCDFWrite(NetCDFWriteUgrid, IOWrite):
                 The dataset dimensions of the data.
 
             bounds: `bool`
-                Whether or not the data represent bounds.
+                Whether or not the data represent cell bounds. This is
+                used to inform the setting of chunksizes for the
+                bounds of 1-d coordinates.
 
                 .. versionadded:: (cfdm) NEXTVERSION
 
@@ -6541,7 +6543,7 @@ class NetCDFWrite(NetCDFWriteUgrid, IOWrite):
             dataset_chunks = chunksizes
         elif chunksizes is not None:
             # Chunked as defined by the tuple of int given by 'data'
-            chunksizes = self._minimum_chunksizes(data, chunksizes, bounds)
+            chunksizes = self._one_d_chunks(data, chunksizes, bounds)
             return False, chunksizes, shards
 
         # Still here? Then work out the chunking strategy from the
@@ -6569,7 +6571,7 @@ class NetCDFWrite(NetCDFWriteUgrid, IOWrite):
             # (250, 250, 4)). However, we only want one number per
             # dimension, so we choose the largest: [96, 250].
             chunksizes = [max(c) for c in chunksizes]
-            chunksizes = self._minimum_chunksizes(data, chunksizes, bounds)
+            chunksizes = self._one_d_chunks(data, chunksizes, bounds)
             return False, chunksizes, shards
         else:
             # The data is scalar, so 'chunksizes' is () => write the
@@ -7534,10 +7536,10 @@ class NetCDFWrite(NetCDFWriteUgrid, IOWrite):
 
         return array
 
-    def _minimum_chunksizes(self, data, chunksizes, bounds=False):
-        """Set minimum chunksizes for particular forms of data.
+    def _one_d_chunks(self, data, chunksizes, bounds=False):
+        """Modify the chunksizes for 1-d data.
 
-        Return the chunksizes unchanged when
+        Return the input *chunksizes* unchanged when
 
         * The data is 0-d.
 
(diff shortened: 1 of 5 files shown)


cfdm/read_write/write.py
```diff
@@ -677,6 +680,46 @@ class write(ReadWrite):
...
+
+              A floating point value is rounded down to the nearest
+              integer, and a string represents a quantity of byte
+              units. For instance a chunksize of 1024 bytes may be
+              specified with any of ``1024``, ``1024.9``, ``'1024'``,
+              ``'1024.9'``, ``'1024 B'``, ``'1 KiB'``, ``'0.001024
+              MB'``, etc. Recognised byte units are (case
+              insensitive): ``B``, ``KiB``, ``MiB``, ``GiB``, ``TiB``,
+              ``PiB``, ``KB``, ``MB``, ``GB``, ``TB``, and
+              ``PB``. Spaces in strings are optional.
```
sadielbartholomew: This documenting of the byte value parsing is certainly clear enough, but I'd suggest also referencing that the behaviour is as per `dask.utils.parse_bytes`, and inter-linking to https://docs.dask.org/en/latest/api.html#dask.utils.parse_bytes if possible, so that users are aware that it is the same processing and can consult explicit examples given there. Specifically I would update this to (also breaking up the longer paragraph to split between the description and the text examples):

```suggestion
              Values are parsed as bytes with optional units as
              accepted by `dask.utils.parse_bytes`, where
              a string represents a quantity of byte
              units and a floating point value is rounded down
              to the nearest integer.
              
              For instance a chunksize of 1024 bytes may be
              specified with any of ``1024``, ``1024.9``, ``'1024'``,
              ``'1024.9'``, ``'1024 B'``, ``'1 KiB'``, ``'0.001024
              MB'``, etc. Recognised byte units are (case
              insensitive): ``B``, ``KiB``, ``MiB``, ``GiB``, ``TiB``,
              ``PiB``, ``KB``, ``MB``, ``GB``, ``TB``, and
              ``PB``. Spaces in strings are optional.
```

But not a blocker so up to you!

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -7499,3 +7535,75 @@ def _pack_array(self, array, scale_factor=None, add_offset=None):
             array = array.astype(dtype)
 
         return array
+
+    def _one_d_chunks(self, data, chunksizes, bounds=False):
+        """Modify the chunksizes for 1-d data.
+
+        Return the input *chunksizes* unchanged when
```
```suggestion
        Return the input *chunksizes* unchanged when:
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -7499,3 +7535,75 @@ def _pack_array(self, array, scale_factor=None, add_offset=None):
...
+
+        * The data is 0-d.
+
+        * The data is not 1-d when *bounds* is False.
+
+        * The data is not 2-d when *bounds* is True.
+
+        * The input chunksize is larger the minimum 1-d chunksize.
```
```suggestion
        * The input chunksize is larger than the minimum 1-d chunksize.
```

although while I am suggesting on this line, to be precise I would make it:


```suggestion
        * The input chunksize is larger than or equal to the minimum
          1-d chunksize.
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -7499,3 +7535,75 @@ def _pack_array(self, array, scale_factor=None, add_offset=None):
...
+
+        For 1-d data (and 2-d data when *bounds* is True), when the input
+        chunk size is less than "one_d_chunks":
+
+        * If the total data size is less than the minimum 1-d chunk
+          size (given by `write_vars['one_d_chunks']`) then set the
```
sadielbartholomew: For consistency with the rest of the docs(tring):


```suggestion
        For 1-d data (and 2-d data when *bounds* is True), when the input
        chunksize is less than "one_d_chunks":

        * If the total data size is less than the minimum 1-d chunksize
          (given by `write_vars['one_d_chunks']`) then set the
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -7499,3 +7535,75 @@ def _pack_array(self, array, scale_factor=None, add_offset=None):
...
+
+        .. versionadded:: (cfdm) NEXTVERSION
+
+        :Parameters:
+
+            data: `Data`
+                The data to be chunked.
+
+            chunksizes: `tuple`
+                The original chunksizes for the data
```
```suggestion
                The original chunksizes for the data.
```

sadielbartholomew (approved):
Lovely! I have only some very minor docs-related suggestions. Please consider those and then merge at will.

cfdm/read_write/netcdf/netcdfwrite.py
davidhassell: The behaviour in the code is strictly larger than, so I think the docstring should stand - it may be that the chunk _shape_ (if not it's size) is changed even when there's equality.

cfdm/read_write/write.py
davidhassell: Nice

davidhassell:
Wonderful. I've improved the `dataset_chunks` docstring along the same lines as your suggestion for `one_d_chunks`.

cfdm/read_write/netcdf/netcdfwrite.py
sadielbartholomew: Ah I see, thanks for clarifying. All good in that case.

PR #416, opened 2026-07-30
API reference: detect missing class pages & populate all gaps

sadielbartholomew:
Sister PR to NCAS-CMS/cf-python#959, doing the equivalent thing for cfdm, namely detecting undocumented classes in the `docs/source/check_docs_api_coverage.py` logic and adding in any missing pages and associated methods. There were 33 (of 77 total, at present) missing in this case!

Note as with the cf-python PR I haven't, for the newly-documented, previously missing classes, organised their methods under sub-headings, because this will take quite a lot of time and I think it is better we get them into the documented first and then we can think about organising them later. For now I've bunged them all under a 'methods' heading and they are in ASCIIbetical order from copying and pasting from the checking script output.


davidhassell (approved):
Hi Sadie - the method and execution all look good to me (I've built the docs to have a look). Please merge when you're ready.

sadielbartholomew:
Thanks for the review David. Noting that the linting CI job is failing for unrelated reasons (some required services are down), merging.

PR #420, opened 2026-08-03
Automatically create consolidated HDF5 metadata with `cfdm.write`

### LINKED ISSUE #413: Automatically create consolidated HDF5 metadata with `cfdm.write`
When reading a netCDF4 dataset, the performance is _hugely_ improved if the HDF5 internal metadata is consolidated into a single block at the start of the file. This doesn't happen by default, because the default HD5 metadata block size is 4096 bytes  - far too small to contain all of the B-tree information. 

However, when we write out to disk with `cfdm.write`, we can analyse the fields being written and estimate a metadata block size that will contain the entire HDF5 B-tree, and when the backend is `h5netcd-h5py`, we can set this metadata block size during the dataset.

davidhassell:
Fixes #413


cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -6375,12 +6508,53 @@ def _file_io_iteration(
...
 
+        if is_log_level_debug(logger):
+            logger.debug("WRITE OPERATIONS:")  # pragma: no cover
+            for method, kwargs in g["write_operations"]:
+                logger.debug(f"  {method}: {kwargs}\n")  # pragma: no cover
+
+        # for method, kwargs in g["write_operations"]:
+        #     print(f"  {method}: {kwargs}\n")
+
+        if (
```
sadielbartholomew, line 6521: To tidy:

```suggestion

        if (
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -6375,12 +6508,53 @@ def _file_io_iteration(
...
+            g["hdf5_consolidated_metadata"]
+            and g["backend"] == "h5netcdf-h5py"
+            and "meta_block_size" not in g["h5py_options"]
+        ):
+            # ---------------------------------------------------------
+            # Calculate a HDF5 meta block size that should be large
+            # enough for all internal file metadata.
+            #
+            # The size is based on the list of write operations
+            # (write_vars['write_opertations']) that deine what is
```
```suggestion
            # (write_vars['write_operations']) that define what is
```

cfdm/read_write/write.py
```diff
@@ -909,6 +909,10 @@ class write(ReadWrite):
             https://docs.h5py.org/en/stable/high/file.html#h5py.File
             for details on which keywords are available.
 
+            .. note:: If the ``'meta_block_size'`` is not provided by
+                      _h5py_options*, then it will automatically set
+                      to a value that
```
sadielbartholomew, line 914: Looks like this still needs finishing:

```suggestion
            .. note:: If the ``'meta_block_size'`` is not provided by
                      *h5py_options*, then it will automatically set
                      to a value that
```

cfdm/read_write/write.py
```diff
@@ -918,6 +922,54 @@ class write(ReadWrite):
 
             .. versionadded:: (cfdm) 1.13.1.0
 
+        hdf5_consolidated_metadata: `bool`, optional
+            If True (the default) then create a single consolidated
+            metadata block in an output HDF5 dataset that should be
+            large enough for the entire B-tree. Such a consolidated
+            HDF5 metadata block gives huge performance benefits when
```
sadielbartholomew: Maybe tweak to this, just to take care we don't over-promise:

```suggestion
            HDF5 metadata block gives significant performance benefits when
```

cfdm/read_write/write.py
```diff
@@ -918,6 +922,54 @@ class write(ReadWrite):
...
+
+        hdf5_expansion_factor: number, optional
+            Configure the algorithm for calculating the size of the
+            metadata block in an output HDF5 dataset that should be
+            large enough for the entire B-tree. Ignored unless
+            *hdf5_consolidated_metadata* is True, the *backend* is
+            ``h5netcdf-h5py`` and ``'meta_block_size'`` has not been
+            provided in the *h5py_options*.
+
+            Such a consolidated HDF5 metadata block gives huge
```
sadielbartholomew: As above suggestion:

```suggestion
            Such a consolidated HDF5 metadata block gives significant
```

cfdm/read_write/write.py
```diff
@@ -918,6 +922,54 @@ class write(ReadWrite):
...
+            performance benefits when reading the dataset. The
+            algorithm that determines this size does so by calculating
+            an approximate amount of bytes based on the actual groups,
+            attributes, variables, and dimension that are going to be
+            written to the dataset; and then multipling this amount by
+            the *hdf5_expansion_factor* to account for the HDF5's
+            dynamic, non-linear memory allocation behaviours, which
+            are not be known in advance. The default value is
```
```suggestion
            are not known in advance. The default value is
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -561,20 +619,43 @@ def _string_length_dimension(self, size):
...
+                    self._createDimension(
+                        group=parent_group, ncdim=ncdim, size=size
+                    )
                 except RuntimeError:
                     pass  # TODO convert to 'raise' via fixes upstream
 
         return ncdim
 
-    def _createDimension(self, group, ncdim, size):
+    def _createDimension(self, **kwargs):  # group, ncdim, size):
```
sadielbartholomew: Tidying:

```suggestion
    def _createDimension(self, **kwargs):
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -336,7 +371,30 @@ def _write_variable_attributes(
 
         return netcdf_attrs
 
-    def _set_attributes(self, attributes, ncvar=None, group=None):
+    def _set_attributes(
+        self, **kwargs
+    ):  # attributes, ncvar=None, group=None):
```
```suggestion
    ):
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -2667,6 +2761,27 @@ def _create_external(
     def _createVariable(self, **kwargs):
         """Create a variable in the dataset.
 
+        .. versionadded:: (cfdm) 1.7.0
+
+        :Parameters:
+
+            kwargs:
+                The same keywords arguments as for `_createVariable_2'.
```
```suggestion
                The same keyword arguments as for `_createVariable_2'.
```

cfdm/read_write/netcdf/netcdfwrite.py
```diff
@@ -3538,8 +3653,26 @@ def _transform_strings(self, data, ncdimensions):
...
         return data, ncdimensions
 
-    def _write_data(
+    def _write_data(self, **kwargs):
+        """Write a data array to the dataset.
+
+        :Parameters:
+
+            kwargs:
+                The same keywords arguments as for `_write_data_2`.
```
```suggestion
                The same keyword arguments as for `_write_data_2`.
```

cfdm/read_write/netcdf/netcdfwrite_meta_block_size.py
```diff
@@ -0,0 +1,185 @@
...
+
+        shape: `tuple` of `int`
+            The shape of the data array
+
+        contiguous: `bool`` or sequence of `int`
+            True if the data array is contigious.
+
+        chunksizes: `None` or sequence of `int`
+            The shape of the data array chunks. Ignored if the data
+            array is contigious (in which case it may be `None`).
```
sadielbartholomew:
```suggestion
        shape: `tuple` of `int`
            The shape of the data array.

        contiguous: `bool` or sequence of `int`
            True if the data array is contiguous.

        chunksizes: `None` or sequence of `int`
            The shape of the data array chunks. Ignored if the data
            array is contiguous (in which case it may be `None`).
```

cfdm/read_write/netcdf/netcdfwrite_meta_block_size.py
```diff
@@ -0,0 +1,185 @@
...
+    Accounts for data chunks.
+
+    .. versionadded: (cfdm) NEXTVERSION
+
+    :Parameters:
+
+        shape: `tuple` of `int`
+            The shape of the data array
+
+        contiguous: `bool`` or sequence of `int`
```
sadielbartholomew:
```suggestion
        contiguous: `bool`
```

The logic doesn't seem to support the sequence case, so I think the docs need amending here.

cfdm/read_write/netcdf/netcdfwrite_meta_block_size.py
```diff
@@ -0,0 +1,185 @@
...
+                    if m == "_createVariable_2"
+                    and ncdim in k.get("dimensions", ())
+                )
+                size = 224 + 16 * n_referencing_vars
+                meta_block_dimensions += size
+
+            elif method == "_createGroup_2":
+                # Meta block space for a group
+                size = 2048 + 40 + len(kwargs["group_name"])
+                meta_block_groups += size
```
sadielbartholomew: Since all the `ifs` concern `method` values, better to `match`/`case`:

```suggestion
            match method:
                case "_set_attributes_2":
                    for name, value in kwargs["attributes"].items():
                        # Meta block space for an attribute
                        size = 51 + len(name.encode("utf-8"))
                        try:
                            size += len(value.encode("utf-8"))
                        except AttributeError:
                            size += np.asanyarray(value).nbytes

                        meta_block_attributes += size

                case "_createVariable_2":
                    n_variables += 1

                    # Meta block space for a variable
                    size = (
                        80
                        + len(kwargs["varname"].encode("utf-8"))
                        + 8 * len(kwargs.get("dimensions", ()))
                    )
                    size += _calculate_chunk_metadata(
                        kwargs.get("shape", ()),
                        kwargs.get("contiguous", True),
                        kwargs.get("chunksizes"),
                    )

                    meta_block_chunk_metadata += size

                case "_createDimension_2":
                    # Meta block space for a dimension
                    ncdim = kwargs["ncdim"]
                    n_referencing_vars = sum(
                        1
                        for m, k in g["write_operations"]
                        if m == "_createVariable_2"
                        and ncdim in k.get("dimensions", ())
                    )
                    size = 224 + 16 * n_referencing_vars
                    meta_block_dimensions += size

                case "_createGroup_2":
                    # Meta block space for a group
                    size = 2048 + 40 + len(kwargs["group_name"])
                    meta_block_groups += size
```

(with an optional `case _` `pass` case if you prefer it that way).

sadielbartholomew (approved):
Looks good, so approving! Though I have a few general comments:

* Ideally we can rename, else document, the `_2` methods (consistently) to indicate that they are the execution methods corresponding to the stored write operations. Maybe renaming from `*_2` to `*_implementation` or `*_impl` or similar (or alternatively drop the `_2` and append onto the other methods with the same name `*_store` or `_process` similar)? I know they are internal but just to make it clear for future reference - e.g. will be useful when we eventually do the `netcdfwrite` refactor!

* No unit test has been added for the new `hdf5_expansion_factor` keyword - is there a reason? Otherwise we should add one in, even if basic. Else raise a follow-on Issue to do this.

Please let me know your thoughts before we look towards merging.

cfdm/read_write/netcdf/netcdfwrite_meta_block_size.py
davidhassell: Nice matching. No need for a "case _: pass" for me :)

davidhassell:
> * Ideally we can rename, else document, the _2 methods (consistently) to indicate that they are the execution methods corresponding to the stored write operations. Maybe renaming from *_2 to *_implementation or *_impl or similar (or alternatively drop the _2 and append onto the other methods with the same name *_store or _process similar)? I know they are internal but just to make it clear for future reference - e.g. will be useful when we eventually do the netcdfwrite refactor!

Good idea. I opted for `_work`, since "implementation" already means something else in the context of this class (i.e. the CF data model implementation). What do you think? https://github.com/NCAS-CMS/cfdm/pull/420/commits/f1623cd212bdc474a389d37f8f0cd2afcd25cdd8

 > * No unit test has been added for the new hdf5_expansion_factor keyword - is there a reason? Otherwise we should add one in, even if basic. Else raise a follow-on Issue to do this.

Fixed: https://github.com/NCAS-CMS/cfdm/pull/420/commits/df5ec4e403d6075870d17e84c9d6db25872ed226

sadielbartholomew:
> What do you think? https://github.com/NCAS-CMS/cfdm/commit/f1623cd212bdc474a389d37f8f0cd2afcd25cdd8

Concise and descriptive, so excellent choice.

> Fixed: https://github.com/NCAS-CMS/cfdm/commit/df5ec4e403d6075870d17e84c9d6db25872ed226

Brill, thanks.

Just doing a final sanity check then it looks like it is ready to merge!

sadielbartholomew (approved):
Thanks for addressing all feedback. Sanity check delayed by lunch but have re-run the tests and eyeballed the new commits and everything looks great. Please merge.

PR #422, opened 2026-08-13
Improve `cfdm.read` performance for UGRID datasets

### LINKED ISSUE #421: Improve `cfdm.read` performance for UGRID datasets
Reading UGRID dataset with `cfdm.read` can be very slow. This is because the Point topology is always calculated non-lazily, even when it is not need by any of the out Fields or Domain. This happens because the shape of the Point Topology is at first unknown (fine!), but it is realised in order to see if we want to store the dataset chunking strategy (we don't!), and the act of realising the shape triggers the full calculation of the topology. This is especially bad because this calculation is very badly coded (e.g. https://github.com/NCAS-CMS/cfdm/blob/v1.13.2.1/cfdm/data/subarray/mixin/pointtopology.py#L81 says `# WARNING (TODO): This loop is a potential performance bottleneck.`)

The solution is two-fold:

1. Stop cfdm.read from trying to store the dataset chunking strategy when the shape of the data is unknown.
2. Re-formulate the badly written code (which loops round nodes in a `for` loop) to use `numpy`. 

#### Use of LLM

The re-formulation into `numpy` was heavily aided by Gemini 3.6 Flash

davidhassell:
Fixes #421


cfdm/read_write/netcdf/netcdfread.py
```diff
@@ -7438,27 +7447,44 @@ def _create_Data(
...
-                shards = self._get_dataset_shards(ncvar)
-                shape = g["variables"][ncvar].shape
-                if shards is not None and shape == data.shape:
-                    self.implementation.nc_set_dataset_shards(data, shards)
+            store_chunks = True
+            try:
+                # Don't store the chunks of a `Data` object that
+                # doesn't know its own shape (i.e. the shape contains
+                # one or more NaNs), because this means that it's
+                # array is not taken directly from an variable in the
```
sadielbartholomew:
```suggestion
                # one or more NaNs), because this means that its
                # array is not taken directly from a variable in the
```

sadielbartholomew (approved):
Great stuff! I also see much improvement in read-in speed for UGRID datasets on this branch, as per the goal. Even with the fairly humble `ugrid_global_2.nc` file from the cf-python test suite it is about 3 times as fast to read in with these changes 🚀 

Please merge after the typo fix suggestion.

PR #428, opened 2026-09-01
Improve `cfdm.environment` (sort, better test) & update docs

sadielbartholomew:
Went in to do some minor cfdm docs updates, but ended up bundling in a bit more. Overall:

* remove reference to the deprecated `NetCDF4Array` being the up-to-date API usage and update some other docstrings;
* add a missing dependency to the `environment()` function (`uritools`);
* whilst I was updating `environment()` thought it would be useful to sort it asciibetically so it is easier to find a given library amongst the output, now we have so many dependencies to report on (will do the same for cf-python's equivalent unless there's any objection)...
* ...and update the testing to ensure we don't forget to add any dependencies as they are added.

### DIFF
--- a/cfdm/functions.py
+++ b/cfdm/functions.py
@@ -458,6 +458,7 @@ def environment(display=True, paths=True):
         "cftime": _get_module_info("cftime"),
         "cfunits": _get_module_info("cfunits"),
         "cfdm": (__version__, os.path.abspath(__file__)),
+        "uritools": _get_module_info("uritools", try_except=True),
     }
     string = "{0}: {1!s}"
     if paths:
@@ -470,6 +471,12 @@ def environment(display=True, paths=True):
             for dep, info in dependency_version_paths_mapping.items()
         ]
     )
+
+    # There are a lot of dependencies (compulsory + optional) to print so
+    # show them in asciibetical order, with the happy coincidence that the
+    # 'cf*' libraries, 'Platform' and 'Python' still come near the top (the
+    # (latter because of the capitalisation) so are easy to pick out.
+    out.sort()
 
     if display:
         print("\n".join(out))  # pragma: no cover
--- a/cfdm/abstract/implementation.py
+++ b/cfdm/abstract/implementation.py
@@ -35,8 +35,12 @@ class Implementation(metaclass=abc.ABCMeta):
 
         >>> i = cfdm.implementation()  # child CF data model implementation
         >>> sorted(i.classes())
-        ['AuxiliaryCoordinate',
+        ['AggregatedArray',
+         'AuxiliaryCoordinate',
          'Bounds',
+         'BoundsFromNodesArray',
+         'CellConnectivity',
+         'CellConnectivityArray',
          'CellMeasure',
          'CellMethod',
          'CoordinateConversion',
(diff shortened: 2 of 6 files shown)


sadielbartholomew:
(Linting CI job is failing because it seems that here's some issue with pre-commit right now:

>Our services aren't available right now ... We're working to restore all services as soon as possible

Though I'm overdue to look at the workflows again and see if anything needs updating so I'll do that soon.)

cfdm/functions.py
```diff
@@ -471,6 +472,12 @@ def environment(display=True, paths=True):
         ]
     )
 
+    # There are a lot of dependencies (compulsory + optional) to print so
+    # show them in asciibetical order, with the happy coincidence that the
+    # 'cf*' libraries, 'Platform' and 'Python' still come near the top (the
+    # (latter because of the capitalisation) so are easy to pick out.
+    out.sort()
```
davidhassell, line 479: For me, however, having `cfdm` coming at the bottom is the most useful, as that is almost always what I'm looking for when I run `cfdm.environment()`, and I don't have to scan a list to find it.

However, logically grouping the dependencies is hard, and asciibetical sorting is certainly natural, so let's do this, and I'm sure I'll adapt :)

cfdm/functions.py
```diff
@@ -458,6 +458,7 @@ def environment(display=True, paths=True):
         "cftime": _get_module_info("cftime"),
         "cfunits": _get_module_info("cfunits"),
         "cfdm": (__version__, os.path.abspath(__file__)),
+        "uritools": _get_module_info("uritools", try_except=True),
     }
```
davidhassell, line 462: Can we add `numpy`, too? (I don't why it was never there!)

cfdm/functions.py
davidhassell, line 437: I don't think we need this anymore, do we?

```suggestion
```

davidhassell (changes requested):
Hi Sadie,

Looks good! Just a couple of extra suggestions which could be included ...

cfdm/functions.py
sadielbartholomew: Good point. I agree there is a good case to highlight cfdm as special because it's the library itself and not a dependency.

> so let's do this, and I'm sure I'll adapt

Yes, though let's revisit this one in future - perhaps we can force `cfdm` and `cfdm.core` to the top at some point. I'll leave this comment 'unresolved' as a reminder.

cfdm/functions.py
sadielbartholomew: Ah don't worry it does get included! But via the inclusion of the `cfdm.core` equivalent function entries:

https://github.com/NCAS-CMS/cfdm/blob/33c45f28f71967067134ddaaaa75393aa2f44ea5/cfdm/functions.py#L439-L440

which contains the fundamentals:

https://github.com/NCAS-CMS/cfdm/blob/33c45f28f71967067134ddaaaa75393aa2f44ea5/cfdm/core/functions.py#L53-L62

So everything is there - indeed my new testing would detect if it wasn't reported on, hence why it is useful.

cfdm/functions.py
sadielbartholomew: Do you mean `import netCDF4` or the empty line? I would take your suggestion directly but the comment I would guess refers to an import line? If the former, both of the imports are still used so unless you can see a way to avoid calling them I'd say we do need them.

davidhassell: Ah - I meant the `import netCDF4`

sadielbartholomew: Still required as discussed externally so marking as resolved.

sadielbartholomew:
Thanks for your comments @davidhassell. I've replied to each but in short I don't think anything needs updating. Please mark as approved to let me know you agree and then I will merge.

davidhassell (approved):
Great - thanks for addressing my questions. Good to go, from my end.

